PR #1, opened 2025-09-26
Use semantic line breaks

clbarnes:
Roughly guided by the spec here https://sembr.org/

This PR contains no changes to the text other than whitespace (and removing a couple of errant apostrophes).

This is obviously a very bikeshed-prone topic; all I'll say is that I think the benefits far outweigh the costs, and also that if such a change is to be made, it must be now, while no other PRs target this repo which would need rebasing.

## Pros

- Plays much more nicely with version control as single units of thought are added, removed, moved, or updated
- Editors don't waste time reflowing the whole paragraph when inserting text into the middle

## Cons

- Right hand edge is more ragged
- Some lines are longer than some editors' width
   - Users are welcome to configure their editors for soft-wrapping
   - It can be useful to be reminded not to write clauses which are too long

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -38,35 +38,36 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL 
 
 <p>
 <dfn>Transitional</dfn> metadata is added to the specification with the
-intention of removing it in the future. Implementations may be expected (MUST) or
-encouraged (SHOULD) to support the reading of the data, but writing will usually
-be optional (MAY). Examples of transitional metadata include custom additions by
-implementations that are later submitted as a formal specification. (See [[#bf2raw]])
+intention of removing it in the future.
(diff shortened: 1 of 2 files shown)


jo-mueller:
Hi @clbarnes , sorry for the late reply. I think until we really accept changes to the spec text itself here, there are some decisions for the formatting to be made (see #3) that turns a lot of the formatting upside down but of which I hope makes some things easier in the long run. Maybe this already takes care of some of the formatting issues you raise? 😬

jo-mueller:
Hi @clbarnes , just played around with it a bit. Could I convince you to rebase the PR on [this branch](https://github.com/ome/ngff-spec/tree/configure-joint-webpage-layout-for-spec/examples/schemas)? Would make a good addition to the overhaul of the spec text in markdown if we decide for it :)

PR #2, opened 2025-09-26
Use a keepachangelog CHANGELOG.md

clbarnes:
The current HTML table means hand-writing HTML and discourages detailed descriptions of every change made in a release.
The markdown form is easier to write and maintain, keeps track of unreleased changes, and more easily supports multiple changes within a version.

I've gone with the format described here https://keepachangelog.com/en/1.0.0/ but I'm not married to the Added/Changed/Removed subsections; could use a free list for author ease.

This is currently a draft because it uses an `include` directive, which behaves differently if the spec is at the top level or not (path must be relative to the root spec).
This is a top-level spec in the context of this repo, but once it's submoduled into  ome/ngff it might not be.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,95 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Added
+
+- Coordinate systems
+
+### Changed
+
+- `coordinateTransformations` metadata
+
+## [0.5.2] - 2025-01-10
+
+### Changed
+
+- Clarified that the `dimension_names` field in `axes` MUST be included
+
+## [0.5.1] - 2025-01-10
+
+### Fixed
+
+- Re-added the improved omero description from PR-191
+
+## [0.5.0] - 2024-11-21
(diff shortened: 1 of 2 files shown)


jo-mueller:
Hi @clbarnes similar to #1, there is a decision to be made for the versioning...If we *really* go down the road of having simple tagged releases of the spec through this repo, then the changelog and versioning of the spec document should simply be taken care of by github/readthedocs

It would be nice though, to autogenerate a changelog for the generated webpages 👀 Is there a possibility to bring these changes here over to #3?

clbarnes:
I think auto-generated changelogs will always lack something compared to what humans _can_ write (whether they _do_ is another matter...). The important thing is that someone who has implemented v0.5 can go down a simple checklist of items in order to support v0.6. I think commits are too granular and PRs may be not descriptive enough for this.

The main idea here is that this changelog format is more readable and more amenable to multiple changes than the HTML.

jo-mueller:
Tbh, I think a list of merged PRs would be alright-ish. True, Commits are overkill to list and we'd have to make sure that PR titles are informative enough, but that can be done. If changes are too major to fit in a PR title, then it's likely that we are looking at an RFC implementation - in that cases, the PR should contain a reference to the RFC text on the main page, where people will find all necessary information about the change.

joshmoore:
It may be overkill for spec development, but what @dstansby put in place for zarr in https://github.com/zarr-developers/zarr-python/pull/2680 has been a big win.

PR #17, opened 2025-10-21
RFC5: coordinate systems and transformations

jo-mueller:
Built on top top of #14 and #20 (requires these two to be merged first). Contains the [changes proposed in the latest versions of rfc5](https://github.com/ome/ngff/pull/350).


will-moore (commented):
Trying to use this branch from the validator...
The `image.schema` has

```
"$id": "https://ngff.openmicroscopy.org/0.6.dev1/schemas/image.schema",
```
and 

```
"$ref": "https://ngff.openmicroscopy.org/0.6.dev1/schemas/_version.schema"
```

jo-mueller:
@dstansby 

> sorry, don't think I'll have the time to look into this in detail. What I will do at some point is try and validate metadata produced by ome-zarr-models against these as a way of checking them (and checking ome-zarr-models!)

That's what I had in mind. Thanks! We can also do that at the Hackathon?

@will-moore 

>The schemas on this branch are currently being used by ome-ngff-validator at https://github.com/ome/ome-ngff-validator/pull/48

Sorry, I probably broke that. I wanted to have everything in one place here to avoid polluting the commit history any worse than it already is :D Changing the schemas for the validator to this place should fix it, though!

will-moore:
@jo-mueller I pushed a fix about the same time as I made that comment (when I realised it was broken).
https://github.com/ome/ome-ngff-validator/pull/48/commits/5afaab5473e3f159d466286703d58778d7104827
So it should be working now?

clbarnes (commented):
Text looks good to me - changes discussed at the hackathon are presumably going into dev3 (remove inverseOf, restructure byDimension)?

The test suites are still using the `0.6dev2` version scheme, which needs updating.

PR #26, opened 2025-11-14
Split tests

clbarnes:
- Refactor test cases to one file per test
- Factor into two types of test
  - "attributes" which represents single attributes object from a single zarr.json
  - "zarr" which represents a zarr hierarchy in a filesystem store (with no chunks)
- Provide a CLI tool which feeds these test cases into a simple "dingus" CLI of the implementor's choice

Example usage:

```sh
git clone https://github.com/clbarnes/ngff-spec.git
ngff-spec/ome_zarr_conformance.py attributes -- path/to/my/dingus --dingus-arg
```

This will iterate over all attributes tests (concurrently) and, with the test path $TEST_PATH, calls `path/to/my/dingus --dingus-arg $TEST_PATH`.

The dingus is expected to read the test from the given path and return `{"valid": true}` if it believes the test data is valid (`false` if not; optionally `"message": str` with more details). Prints out a TSV of all test names and their status `pass`/`fail`/`error`.

In "attributes" mode, $TEST_PATH is a path to a JSON file containing the Zarr attributes object (i.e. `{"ome": {...}}`). In "zarr" mode, $TEST_PATH is a path to directory representing the root of a zarr container (i.e. containing a `zarr.json`).

Test cases are not guaranteed to be correct for the moment.

I recommend ignoring most of the diff (mainly just moving around JSON files); look at `ome_zarr_conformance.py` and the changes in the `schemas/` directory.


jo-mueller:
3cts from me:

- Version here is `0.6.dev1` but I think in the future (and elsewhere already for `0.6dev2`) we/I am using `0.6dev2` instead of `0.6.dev2`.
- Was all of this done via code? If so - could you point that code at #17? I already combed through all the examples there and reformatted. The pytest tests are also correctly triggered there, so maybe that would also be a good place "for a fresh start"?
- Maybe move the `strict` test cases into `valid`/`invalid`? Not a must, though. At the end, the point is more to be able to pinpoint a failing test to a particular file, anyway, right?

*Edit*: PS: dingus is a much better example placeholder than that foo/bar scheme.

PR #27, opened 2025-11-18
Set BSD3 license

jo-mueller:
Title. Sets the license of the spec text both in the myst metadata and in the repo frontpage as a `LICENSE` file.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,27 @@
+Copyright (C) 2020 Open Microscopy Environment:
+  - Glencoe Software, Inc.
+  - University of Dundee
+  - German BioImaging e.V.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
+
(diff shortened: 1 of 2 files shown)


dstansby:
Given much of the content here is derived from https://github.com/ome/ngff, which was made available (at the time it was copied) under https://www.w3.org/copyright/software-license-2015/, the conditions of that license should probably be followed too? In particular, it requires including the following in re-distributions (which this repository effectively is):

- The full text of this NOTICE in a location viewable to users of the redistributed or derivative work.
- Any pre-existing intellectual property disclaimers, notices, or terms and conditions. If none exist, the [W3C Software and Document Short Notice](https://www.w3.org/Consortium/Legal/copyright-software-short-notice) should be included.
- Notice of any changes or modifications, through a copyright statement on the new code or document such as "This software or document includes material copied from or derived from [title and URI of the W3C document]. Copyright © [YEAR] W3C® (MIT, ERCIM, Keio, Beihang)."

PR #44, opened 2026-01-09
update header for development

jo-mueller:
When people start sending PRs to ngff-spec during the ramp-up to 0.6, we should change the header information which currently states:

```{warning}
This is **not** the released version of the ngff-specification. It is an explorative, work-in-progress document.
```

I suggest to update this to:
```{warning}
This is **not** the released version of the ngff-specification.
It is a work-in-progress document.
Upon release, this warning will be removed and the version number in the document updated.
```

### DIFF
--- a/index.md
+++ b/index.md
@@ -11,7 +11,10 @@ short_title: OME-Zarr
 ## Abstract
 
 ```{warning}
-This is **not** the released version of the ngff-specification. It is an explorative, work-in-progress document.
+This is **not** the released version of the ngff-specification.
+It is a work-in-progress document.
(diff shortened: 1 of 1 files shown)


joshmoore:
:+1:

Sidenote:
 * zarr-python v2 used commented sections (https://raw.githubusercontent.com/zarr-developers/zarr-python/refs/heads/support/v2/docs/release.rst) to do this
 * zarr-python v3 updated to towncrier but not sure if it can provide this type of swap (https://github.com/zarr-developers/zarr-python/blob/acdd892234dbb62833c5d57e59b94d180cd1115d/.readthedocs.yaml#L12)
 
 
image

PR #47, opened 2026-01-09
Add title and description to axis schema

dstansby:
I'm hoping this will replace "root items" in the below rendering: 🤞 

Screenshot 2026-01-09 at 13 15 08


dstansby:
😦 it didn't work

PR #61, opened 2026-01-13
 fix 0.4: use the DOI for all versions 

lubianat:
Updated DOI reference in citing.md

Closes https://github.com/ome/ngff/pull/204/

@jo-mueller should I also update the index.bs file? or rather should we just delete it?


jo-mueller:
@lubianat Thanks for taking this one! I think the decision was to keep the bikeshed files for posterity so that the built documents can still be served under `ngff.openmicroscopy.org/{version}/index.html`

PR #65, opened 2026-01-13
fix 0.5: update DOI for all versions

lubianat:


lubianat:
@joshmoore I wasn't 100% sure on what the date here referred to. 

Should we update the date for every release, then?

PR #87, opened 2026-02-05
omero fields optional

### LINKED ISSUE #192: relax omero metadata
### Motivation
With this issue, I would like to discuss whether ngff could relax the omero metadata requirements by making "visualization specific" parameters (["window"](https://github.com/ome/ngff/blob/15565cb8301e14fc9ba2fc63b6332a97534cee67/0.4/schemas/image.schema#L98) and ["color"](https://github.com/ome/ngff/blob/15565cb8301e14fc9ba2fc63b6332a97534cee67/0.4/schemas/image.schema#L91) ) optional.

As discussed in https://github.com/ome/ngff/issues/155 , there seems to be agreement that the channel name information (e.g. in a microscopy image, the name of the fluorescent markers/antibodies such as "DAPI", "Actin" etc.) should be saved in omero metadata. This information would be stored under "label" according to [current spec](https://ngff.openmicroscopy.org/latest/#omero-md). This type of information is critical to be stored especially in highly-multiplexed imaging techniques such as CyCIF, Codex, Imaging Mass Cytometry etc.

Users might want to save the channel name information without specifying "color" or "window" information. Currently, it is not possible, as there are two required arguments by the omero specs which are, imho, purely for "visualization" purposes. Users would therefore have to specify "arbitrary" information on aesthetics that is not necessarily useful/actionable. 

### Potential solution
Make the current required parameters ("color" and "window") as optional. This would enable the use case of only saving channel name information without having to specify the remaining visualization parameters.

will-moore:
Fixes https://github.com/ome/ngff/issues/192

ported from https://github.com/ome/ngff/pull/297/


index.md
```diff
@@ -1524,14 +1524,21 @@ for more information.
...
-Each dictionary in `channels` MUST contain the field `color`,
-which is a string of 6 hexadecimal digits specifying the color of the channel in RGB format.
-Each dictionary in `channels` MUST contain the field `window`,
-which is a dictionary describing the windowing of the channel.
-The field `window` MUST contain the fields `min` and `max`,
-which are the minimum and maximum values of the window, respectively.
-It MUST also contain the fields `start` and `end`,
-which are the start and end values of the window, respectively.
+The `channels` array length SHOULD correspond to the size of the `c` axis.
+Each dictionary in `channels` MAY contain the following fields:
```
jo-mueller:
```suggestion
Each object in `channels` MAY contain the following fields:
```

index.md
```diff
@@ -1524,14 +1524,21 @@ for more information.
...
 which is an array of dictionaries describing the channels of the image.
-Each dictionary in `channels` MUST contain the field `color`,
-which is a string of 6 hexadecimal digits specifying the color of the channel in RGB format.
-Each dictionary in `channels` MUST contain the field `window`,
-which is a dictionary describing the windowing of the channel.
-The field `window` MUST contain the fields `min` and `max`,
-which are the minimum and maximum values of the window, respectively.
-It MUST also contain the fields `start` and `end`,
-which are the start and end values of the window, respectively.
+The `channels` array length SHOULD correspond to the size of the `c` axis.
```
jo-mueller: A `c`-named axis isn't required to be present

```suggestion
The `channels` array length SHOULD correspond to the size of the respective channels axis, if present.
```

lubianat (approved):

jo-mueller (approved):

lubianat (approved):
I am not super familiar with details, but it looks good as far as I can tell.

index.md
```diff
@@ -1365,14 +1366,24 @@ for more information.
...
-It MUST also contain the fields `start` and `end`,
-which are the start and end values of the window, respectively.
+The `channels` array length SHOULD correspond to the size of the respective channels axis, if present.
+Each object in `channels` is optional and MAY contain the following fields:
+
+- `color` (string) String of 6 hexadecimal digits specifying the color of the channel in RGB format.
+- `label` (string) Channel name.
+- `active` (boolean) Indicates whether the channel should be displayed.
+- `coefficient`: (float) Value to multiply the pixel values of the channel by when rendering.
+- `family`: (string) Used transfer function family for rendering the channel, e.g. "linear", "log", "gamma".
```
jburel: this is not correct

version_history.md
```diff
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - For detailed set of changes see [complete overview](https://ngff.openmicroscopy.org/rfc/5/responses/2/index.html)
 - Updated version keys from `0.6.dev2` to `0.6.dev3` everywhere
 - style: Homogeneous use of backticks in spec document
+- spec: Released requirements for `omero` metadata in the spec document and in schemas.
```
jburel: We should be more explicit and use the term ``specification``

jburel (changes requested):

joshmoore:
@will-moore @jburel: would you like to get this into one of the upcoming .dev versions? And if so, when would you be thinking?

will-moore:
I actually thought this had made it into v0.6.
So, yes. As soon as possible, thanks.

joshmoore:
@jni is leading dev1 so I'll defer. Otherwise we can get it on a board for dev2 and discuss.

jni:
@will-moore there's a number of unresolved comments here. (log->logarithmic, removing "gamma", removing "#" from the color). Otherwise, I'm happy for this to go into 0.9.dev1 which is due ~end August. Note also that I don't have merge rights in this repo :sweat_smile: so someone else will have to do so, presumably *after* 0.6 is out. (Although, given the backwards compatibility of the change, I would not be opposed if it went into 0.6.rc1/final. CC @jo-mueller)

index.md
will-moore: I think we should remove the "family" and "coefficient" terms from the "omero" examples and docs.
They have never been defined before, and are still not clear, so I don't think they will have ever been used.

will-moore:
I think that is all the fixes done.
I was checking https://omero.readthedocs.io/en/stable/developers/Web/WebGateway.html#imgdata for any reference to `coefficient` and `family` but it has no info on these and actually has less info than is in these docs themselves so I've removed that link.

I also have therefore also removed the `coefficient` and `family` fields from the `omero` block since these have never been defined and never used AFAIK.

jni:
Great. @will-moore do you want to mark this as ready for review then?

@jo-mueller what do you think about shipping this with 0.6 final?

jo-mueller:
@jni sorry it took me so long to catch up here. To summarize:

| Feature                     | Version 1 (Before)                                                                 | Version 2 (After)                                                                                     |
|-----------------------------|------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------|
| **`omero` Metadata**        | Optional, but if present, `channels` MUST be included.                             | Not explicitly mentioned.                                                                             |
| **`channels` Array**        | MUST exist if `omero` is present.                                                  | Optional. Length SHOULD correspond to the size of the respective channels axis, if present.           |
| **Fields in `channels`**    | Every object MUST contain `color` and `window`.                                    | All fields are optional. MAY include `color`, `label`, `active`, `window`, or `inverted`.             |
| **`color` Field**           | MUST be a 6-digit hexadecimal RGB string.                                          | MAY be a 6-digit hexadecimal RGB string.                                                             |
| **`window` Object**         | MUST contain `min`, `max`, `start`, and `end`.                                     | If provided, MUST contain `start` and `end`. `min` and `max` are optional.                           |
| **Additional Fields**       | None.                                                                              | MAY include `label` (channel name), `active` (display toggle), and `inverted` (pixel rendering).      |
| **Data Types**              | Implied integers for `min`, `max`, `start`, and `end`.                             | Explicitly allows floating-point numbers for `min`, `max`, `start`, and `end`.                       |
| **RFC 2119 Keywords**       | Uses **MUST** for all requirements (strict compliance).                            | Uses **SHOULD**, **MAY**, and **MUST** (flexible, with conditional requirements).                     |
| **Purpose**                 | Strictly defines rendering requirements (`color` and `window`).                    | Expands to include usability, customization, and rendering flexibility.                              |

Arguments for adding:

- Syncs spec with what was already in the schemas (i.e., `label`, `active`, `inverted` have been in the schemas for a while(?)), and this matches the text with them
- All changes are relaxing requirements, so non-breaking

Arguments against:

- timing?

jo-mueller:
After some more thought, since I made a [quite verbal statement](https://forum.image.sc/t/ngff-weekly-dev-update-thread/110810/123?u=johannessoltwedel) about no new changes in 0.6, I'm hesitant to eat from the forbidden fruit and slip in this (very reasonable!) PR 🙈 

I think my headcanon for moving this forward would be:
1. Get the feedback from @jburel - I think there were some unresolved comments further up?
2. I agree that the requirements are unreasonably tight - but writers will probably circumvent this by writing some reasonable defaults, i.e., for `min`/`max` or `start`/`end`. Making these fields non-required would make these defaults a bit more opinionated as they are now. Aside from the fact that validators would need to update code.
3. Get this PR in first thing after the 0.6 tag?

lubianat:
related to the redo of omero fields, here is an insteesting scenario where color is being used as a local ID for a channel

https://github.com/JurgenKriel/macro_candi

*Channels are resolved by colour, not index. canmac/io/channels.py reads the OME-Zarr omero metadata and matches hex colours, so channel order changes don't silently mis-assign data. Edit COLORS there for your fluorophores.*


this touches on both the need of a channel spec and on different uses of the `omero` metadata block

joshmoore:
Changed based branch to 0.9dev for inclusion in the 0.9.dev1

PR #89, opened 2026-02-17
Make image label key mandatory for label images

### LINKED ISSUE #339: No way to distinguish image and labels-image groups in 0.5
In version 0.4, it was possible to tell whether a group was a image group or an image labels group, because an image labels group was required to have a `image-label` entry in the metadata (https://ngff.openmicroscopy.org/0.4/index.html#label-md).

In version 0.5, it is no longer possible to distinguish between these two types of group, because image labels are no longer required to have an `image-label` entry. Instead, the spec only says this entry should (not must) be present. (https://ngff.openmicroscopy.org/0.5/index.html#labels-md). 

This means that every valid image group is also a valid image-label group, and vice versa.

Would it be useful to still distinguish between these two types of group, and if so, introduce some new metadata that allows them to be distinguished?

Personally I think there is value in distinguishing between these groups so they can be interpreted differently by image veiwers (a labels dataset should be displayed differently to an imaging dataset for example) and readers, and for scientists to signal their intent as to whether a group is an image or labels.

dstansby:
Fixes https://github.com/ome/ngff/issues/339 - see discussion on that thread for motivation.


dstansby:
Does anyone from the OME side have thoughts on the next steps here? It would be nice to have a decision on whether this is merged in time for 0.6.

jo-mueller:
Where I am concerned, I'll say 👎 Long story short, I think a label image should not be flagged as an invalid image only because it doesn't declare itself to be a label image properly. That will still mean, that it is a valid image (although the default color settings viewers may choose may not make much sense for it).

A longer version of this line of thought is found in [this issue](https://github.com/ome/ngff/issues/339). Making this declaration more explicit or introduce a sort of typing of images is something I see more on the plate of RFC8, if anywhere.

dstansby:
> a label image should not be flagged as an invalid image only because it doesn't declare itself to be a label image properly.

Without the image-label key, the data will still consitute a valid OME-Zarr image group. So I'm not sure why this is a concern?

jo-mueller:
I would read 

> In addition to the `multiscales` key, the OME-Zarr Metadata in this image-level `zarr.json` file MUST contain another key, `image-label`,

as an equivalent statement to "If it is a label image and doesn't have the `image-label` field present, it is invalid?

dstansby:
If it is a label image and doesn't have the image-label field present, it would be an invalid image label group, but that doesn't prevent it from being a valid image group.

joshmoore:
Suggest we evaluate this in the context of RFC-8

PR #96, opened 2026-03-02
Fix failing `well` tests

jo-mueller:
Fixes https://github.com/ome/ngff-spec/issues/95

In the `invalid` tests, the flag `"valid": false` flag has been missing, which prevents invalid metadata being recognized as such. I also added the version tag for all the test JSON that were added by #71.


joshmoore (approved):
ty 🙌🏽

PR #114, opened 2026-03-23
fix: avoid non-unicode characters (0.3)

jo-mueller:
Re https://github.com/ome/ngff/issues/479


lubianat (changes requested):
footer indentation still weird here  I think

clbarnes (approved):

clbarnes (approved):

clbarnes (approved):

lubianat (approved):

PR #124, opened 2026-04-02
Fix test cases in v0.5

clbarnes:
Fixes the test cases listed in https://github.com/ome/ngff/issues/325

### DIFF
--- a/tests/plate_suite.json
+++ b/tests/plate_suite.json
@@ -22,7 +22,7 @@
             ],
             "wells": [
               {
-                "path": "A/1",
+                "path": "1/A",
                 "rowIndex": 0,
                 "columnIndex": 0
               }
@@ -55,7 +55,7 @@
             ],
             "wells": [
               {
-                "path": "A/1",
+                "path": "1/A",
                 "rowIndex": 0,
                 "columnIndex": 0
               }
@@ -78,7 +78,7 @@
             ],
             "wells": [
               {
-                "path": "A/1",
+                "path": "1/A",
                 "rowIndex": 0,
                 "columnIndex": 0
               }
@@ -102,7 +102,7 @@
             "rows": [],
             "wells": [
               {
-                "path": "A/1",
+                "path": "1/A",
                 "rowIndex": 0,
                 "columnIndex": 0
               }
@@ -133,7 +133,7 @@
             ],
             "wells": [
               {
-                "path": "A/1",
+                "path": "1/A",
                 "rowIndex": 0,
                 "columnIndex": 0
               }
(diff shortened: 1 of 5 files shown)


clbarnes:
The failing test is somewhat subjective, about the image label's color JSON array. The spec says "this contains one entry per label", which is probably a MUST. If the label zarr array were zero-size, the color JSON array could strictly be empty and valid, although that's obviously a pathological case. I'm OK with reverting this test to empty color arrays as invalid.

clbarnes:
If this is acceptable I'll make the same changes to the previous and future versions.

tests/plate_suite.json
```diff
@@ -22,7 +22,7 @@
             ],
             "wells": [
               {
-                "path": "A/1",
+                "path": "1/A",
```
jo-mueller: ...Is that correct though? In the example in the spec it's always row/column, where the row is indexed by letters and the column is indexed by numbers (see example under https://ngff.openmicroscopy.org/specifications/dev/index.html#plate-metadata)

jo-mueller: See also also: https://ome.github.io/ome-ngff-validator/?source=https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0001A/2551.zarr

jo-mueller: Or is this indented to fail validation? In that case the `"valid": true` is confusing me...

jo-mueller (changes requested):

tests/plate_suite.json
clbarnes: From https://ngff.openmicroscopy.org/specifications/dev/index.html#plate-metadata

> The `path` MUST consist of a name in the `rows` array, a file separator (/), and a name from the `columns` array, in that order.

Here there is only one row, with a name `1`, and only one column, with a name `A`; therefore `{row}/{column}` must be `1/A`.

clbarnes: Letter vs number is inconsequential, it's just that it's got to be in the right order. This test (and others) could be fixed by either changing the names of the rows and the columns, or by changing the path of the well. I did the latter but am open to doing the former.

jo-mueller (approved):

jo-mueller:
LGTM 👍

lubianat:
think the fixes are good, but I wonder if we could drop updates for some of the older versions, if only to reduce the noise here and in the dependabot PRs

Following the lead on ome-zarr-py, maybe we don't propagate this (and similar) changes for ≤ 0.3? 

Not sure on the correct protocol here, though

joshmoore:
Leaving @clbarnes to backport to 0.4 and 0.6 (tests didn't exist before then)

PR #131, opened 2026-05-21
Update version history for changes to the labels specification

sbesson:
We came across this inconsistency while trying to establish the expectations in terms of data types for label images in the [OME-Zarr 0.4 specification](https://ngff.openmicroscopy.org/0.4).

Improvements were proposed and integrated in https://github.com/ome/ngff/pull/165 including some clarification on the supported data types. As noted in the [last comment](https://github.com/ome/ngff/pull/165#issuecomment-1582131845) these changes were published in the [version 0.5 of the specification](https://ngff.openmicroscopy.org/0.5/#labels-md) while the version history suggests they were added in the version 0.4.

This proposal rectifies the version history file of the ngff-spec accordingly. While reviewing this, I noted we have various copies of the version history page: one on each branch of this repository as well as sections in the specification pages like https://ngff.openmicroscopy.org/0.4/#history. I am unclear which one is (are) the single source of truths. Happy to port this change to all relevant places once it's been reviewed  & approved.

PR #132, opened 2026-05-22
Update version history for changes to the labels specification

sbesson:
See https://github.com/ome/ngff-spec/pull/131#issuecomment-4507159671


joshmoore:
Thanks, @sbesson!

PR #134, opened 2026-05-26
Simplify coordinate/displacement metadata

### LINKED ISSUE #520: RFC-5: relocate coordinates/displacements index transformations
Stemming from a discussion on Zulip here https://imagesc.zulipchat.com/#narrow/channel/328251-NGFF/topic/RFC-5/near/596975345

Coordinates and displacement transformations store a (possibly large) Zarr array; calculating a coordinate's transformation indexes into this array. For displacements, it is useful for the stored vectors to have the same units as the incoming and/or outgoing coordinates rather than be locked to voxel/ array-index space, which means that we need an additional transformation to convert the incoming (world space) coordinates to array index space. A displacement would then need to be applied to the _original_ input coordinate, not the array index.

Currently (0.6dev4), this is handled by the array's attributes storing an attribute like

```jsonc
{
  "ome": {
    "version": "0.6",
    "coordinateSystems": [
      ... // must match incoming coordinates' system, plus an axis of type "coordinate" or "displacement"
    ],
    "coordinateTransformations": [ 
      ... // must only be scale/transformation/identity, from voxel space to the above system
    ]
  }
}
```

This introduces quite a lot of special casing and complexity, as it is the only place where

- an `ome` object appears in Zarr array metadata
- `coordinateSystems`/`coordinateTransformations` appear in the root of an `ome` object
- an axis of type `coordinate` or `displacement` can be used
- a transformation's `input` is omitted (implicitly the array indices) but `output` must be given
- a transformation is defined which doesn't participate in the transform graph in the same way

If instead we stored the input->array_index transformation in the coordinate/displacement transformation definition, and added a simplifying constraint to the vector fields, we can get rid of all of these special cases. That is:

```jsonc
"coordinateTransformations": [
  {
    "name": "myDisplacement",
    "type": "displacement",
    "input": { "name": "input" },
    "output": { "name": "output" },
    "path": "coordinateSystems/myDisplacement",
    "indexTransformation": {  // <-- NEW FIELD
      // Maps array index space -> input coordinate space.
      // MUST be identity, scale, or sequence[scale, translation].
      "type": "scale",
      "scale": [1.0, 2.0, 3.0]
    }
  }
]
```

Additionally,

```markdown
The input coordinate system has an axes array of length `Di`,
and the output coordinate system has an axes array of length `Do`.

The displacement/coordinate Zarr array MUST have `Di + 1` dimensions.
Dimensions `0...Di` MUST correspond to the axes of the input coordinate in the same order.
The last axis MUST be of length `Do` and represents the displacement or coordinate vector.

In displacement transformations, `Di = Do`.
```

The procedure for applying the coord/disp transformation would then be:

- receive `input` coordinate
- invert the `indexTransformation` and apply the inversion to the input coordinate to get an array index
- use this array index to look up or interpolate a vector from the coord/disp Zarr array
- if `type:coordinate`, that is the `output` coordinate
- if `type:displacement`, add the vector to the `input` coordinate to calculate the `output` coordinate

I am open to discussion on the name of the `indexTransformation` field but feel quite strongly that it shouldn't just be called `transformation`, because it doesn't participate in the transform graph in the same way.

clbarnes:
Currently, locating the vector fields associated with coordinates and displacements transforms in the input coordinate system required an extra coordinate system and coordinate transformation metadata on the vector field Zarr array. This proposal moves it to within the existing transformation metadata, erasing the need for a number of metadata features which were only used for this purpose.

Resolves https://github.com/ome/ngff/issues/520 . Notes about the general issue should go there; preferably keep discussion here to the proposed text/ supporting information changes.

Also changes a few examples to use the conventional `coordinateTransformations/` location for coordinate transformation Zarr nodes.

### DIFF
--- a/examples/transformations/coordinates1d.json
+++ b/examples/transformations/coordinates1d.json
@@ -6,7 +6,10 @@
   "coordinateTransformations": [{
     "name": "a coordinate field transform",
     "type": "coordinates",
-    "path": "i2xCoordinates",
+    "path": "coordinateTransformations/i2xCoordinates",
+    "indexTransformation": {
+      "type": "identity"
+    },
     "input": {"name": "input"},
     "output": {"name": "output"},
     "interpolation": "nearest"
--- a/examples/transformations/displacement1d.json
+++ b/examples/transformations/displacement1d.json
@@ -6,7 +6,14 @@
   "coordinateTransformations": [{
     "name": "a displacement field transform",
     "type": "displacements",
-    "path": "i2xCoordinates",
+    "path": "coordinateTransformations/i2xCoordinates",
+    "indexTransformation": {
+      "type": "scale",
+      "scale": [
+        2,
+        1
+      ]
+    },
     "input": {"name": "input"},
     "output": {"name": "output"},
     "interpolation": "nearest"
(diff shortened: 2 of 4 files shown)


jo-mueller:
Superseded by #145?

clbarnes:
Yes.

For posterity: we decided to turn vector field transforms into multiscale arrays and make use of existing multiscale metadata to handle index transformations.

PR #137, opened 2026-06-10
Streamline input output constraints

### LINKED ISSUE #159: Confusing wording on requirement of path/name in root level multiscale.coordinateTransformations
### Describe the issue

The `multiscales_reference_to_label.json` example has this `ome.multiscales[].coordinateTransformations`

https://github.com/ome/ngff-spec/blob/36f74596b935606bfa53464ec18e808bce4b0531/examples/multiscales_strict/multiscales_reference_to_label.json#L30-L36

but that seems to contradict this part of the prose, that suggests that 

> The `path` field of `input` SHOULD be omitted. [inside of a root level `multiscale.coordinateTransformations[]` item]

https://github.com/ome/ngff-spec/blob/36f74596b935606bfa53464ec18e808bce4b0531/index.md?plain=1#L1350-L1356

....

i would even argue that the prose itself is *almost* self-contradictory when it says:

"The values of both `input` and `output` fields MUST be an object with fields `name` and `path` ..."  and then goes on to say "The `path` field of `input` SHOULD be omitted."

Can someone confirm what the intended semantics are?  Maybe this wording could be improved.

### LINKED ISSUE #523: RFC-5: Type inconsistency on nested transform input/output
Problematic section:

> **Omitting input/output**: Coordinate transformations MUST specify their input and output coordinate systems using the input and output fields. These fields MUST correspond to the name of a coordinate system or the path to a multiscales group. Exceptions are if the coordinate transformation is wrapped in another transformation, e.g. as part of a sequence, byDimension or bijection. In these cases, the input and output fields MAY be omitted or null.

The rest of the spec is now consistent about transform input/outputs always being objects and never plain strings. This paragraph still requires input/output on nested transforms being plain strings, if not omitted.

I've mentioned before that my opinion is that they SHOULD/MUST be omitted when nested, and that would make this paragraph simple too :)

I'd otherwise recommend to require the input/output of nested transforms to follow the same constraints as their parent (i.e. depending on the multiscale/dataset/scene context).

Also, not sure anymore what to do with this issue. I thought it was a minor phrasing issue, and only realised it's a bit more substantially about actual requirements while trying to phrase out the "simple" fix.

I guess the issue remains as the inconsistency in definitions. The first rule we encounter when reading from the top is:

> MUST contain the field output, which is an object with fields name and path. The output field MAY be omitted if the transformation is part of a wrapper transform (i.e., [sequence](https://ngff.openmicroscopy.org/specifications/dev/index.html#sequence-md), [bijection](https://ngff.openmicroscopy.org/specifications/dev/index.html#bijection-md), [byDimension](https://ngff.openmicroscopy.org/specifications/dev/index.html#bydimension-md), see details).
> MUST contain the field input, which is an object with fields name and path. The input field MAY be omitted if the transformation is part of a wrapper transform (i.e., [sequence](https://ngff.openmicroscopy.org/specifications/dev/index.html#sequence-md), [bijection](https://ngff.openmicroscopy.org/specifications/dev/index.html#bijection-md), [byDimension](https://ngff.openmicroscopy.org/specifications/dev/index.html#bydimension-md), see details).

But then as we get down to the "Additional details", the input/output are specified as "MUST be a name/path".

jo-mueller:
Fixes https://github.com/ome/ngff/issues/523
Fixes #159 

- removes some duplicate text and adds more descriptive tables to the transformations section to make the whole paragraph a bit easier to read.
- synchronizes the constraints introduced for transformation input/output fields across the spec so that they are progressively narrowed down from broad to context-dependend tight
- relaxes restriction on directionality in `ms > coordinateTransformations`, which makes the text cleaner and has no practical implication since we are only allowing identity/scale/translate/sequence here.

cc @btbest


index.md
```diff
@@ -403,19 +403,15 @@ Coordinate transforms are in the "forward" direction.
...
-  The `input` field MAY be omitted if the transformation is part of a wrapper transform
-  (i.e., [`sequence`](#sequence-md), [`bijection`](#bijection-md), [`byDimension`](#bydimension-md), see details).
-- MAY contain the field `name` (string).
-  Its value MUST be unique across all `name` fields for all coordinate transformations in the same list.
-- Parameter values MUST be compatible with input and output space dimensionality (see details).
+A transform is a JSON object with the following fields:
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `type` | string | yes | The type of the transformation, which determines the required and optional fields for that transformation. |
```
clbarnes: Suggestion to replace the "required" column with a "requirement level" (or synonym), with MUST/SHOULD/MAY to match spec text.

index.md
```diff
@@ -403,19 +403,15 @@ Coordinate transforms are in the "forward" direction.
...
+| `name` | string | no | A unique name for this transformation. |
+| `input` | object | yes, unless part of a wrapper transform ([`sequence`](#sequence-md), [`bijection`](#bijection-md), [`byDimension`](#bydimension-md)) | The input coordinate system for this transformation. |
```
clbarnes: The Type column, where the type is another object defined in this spec, should use that object's name and link to its definition. But we shouldn't try to change all the definitions into tables in this PR :)

clbarnes (changes requested):

index.md
jo-mueller: It was tempting 😬 🙈

index.md
jo-mueller: Agree....but that's for a different PR. But, since the requirement level is often context-dependent, I'd rather leave it as a yes/no and then further narrow it in the respective context.

index.md
jo-mueller: e876295e17b4ecb4dcb8e88e293d9139f77d2ec5

clbarnes (approved):

index.md
```diff
@@ -403,21 +403,18 @@ Coordinate transforms are in the "forward" direction.
...
-  The `input` field MAY be omitted if the transformation is part of a wrapper transform
-  (i.e., [`sequence`](#sequence-md), [`bijection`](#bijection-md), [`byDimension`](#bydimension-md), see details).
-- MAY contain the field `name` (string).
-  Its value MUST be unique across all `name` fields for all coordinate transformations in the same list.
-- Parameter values MUST be compatible with input and output space dimensionality (see details).
+A transform is a JSON object with the following fields:
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| [`type`](transform-type-md)= | string | yes | The type of the transformation, which determines the required and optional fields for that transformation. Possible values |
```
btbest:
```suggestion
| [`type`](#transform-type-md) | string | yes | The type of the transformation, which determines the required and optional fields for that transformation. Possible values |
```

:)

(Also, what is "Possible values" about)?

jo-mueller: Eh... I think I started writing something and then recondiered. Thanks for the catch 🙈

jo-mueller: 11084a30c5f205e9a9a8f437e13e5a05afeab406

jo-mueller:
Also, I noticed that in #117 , I introduced a constraint that would require all transformations under `multiscale > coordinateTransformations` to have the same directionality, i.e., `"intrinsic"` should be referenced by `input`, and the other should be the `output`, which is unnecessarily strict. I am removing this constraint here, which makes things easier.

cc @btbest @will-moore @clbarnes @tlambert03 @dstansby @kevinyamauchi

btbest:
Happy with loosening the directionality, but the current MAY phrasing accidentally reintroduces the "multiscale transforms can point anywhere" problem:

>  - One of `input` or `output` MUST reference the "intrinsic" coordinate system by `name` (`path` MAY be omitted or null).
>  - The other MAY reference a named coordinate system in the same multiscales group (by `name`, `path` MAY be omitted or null), or in a child [labels](#labels-md) group (by `name` and `path`).
>  - When referencing a child labels group, the transformation MUST be [`identity`](#identity-md), [`scale`](#scale-md), or [`translation`](#translation-md).

The MAY in the second bullet point is the catch. Maybe I just have mischief baked in, but I immediately read this as "Well, you never told me I can't reference some other multiscale or a scene".

This should say "The other MUST reference a coordinate system either in... or in..." (the same goes for line 1328)

jo-mueller:
> This should say "The other MUST reference a coordinate system either in... or in..." (the same goes for line 1328)

Right!

jo-mueller:
@tlambert03 would you mind you have a quick look at this branch and see if you find the wording around the constraints around the `input` and `output` fields and where they apply clearer here? 🙂

index.md
```diff
@@ -432,39 +429,18 @@ The following transformations are supported:
...
-      "type": "scale",
-      "scale": [2, 3.12],
-      "input": {"name": "in"},
-      "output": {"name": "out"}
-    }
-  ]
-}
-
-```
+The parameter values (i.e., `scale` for a [scale transformatiion](#scale-md)) MUST be compatible with input and output space dimensionality (see details).
```
btbest:
```suggestion
The parameter values (e.g., `scale` for a [scale transformatiion](#scale-md)) MUST be compatible with input and output space dimensionality (see details). 
```

index.md
```diff
@@ -484,15 +461,17 @@ Depending on which, different constraints apply to the transformations, as descr
...
-  - The `coordinateTransformations` field MUST contain an array of valid [transformations](#trafo-types-md).
-  - The `input` to every one of these transformations MUST be the same coordinate system, referenced by the `name` field.
-  - The `output` MUST be another coordinate system defined under `multiscales > coordinateSystems`, referenced by the `name` field, or another coordinate system in a child [labels](#labels-md) group referenced by both `name` and `path`.
-- **Inside `scene > coordinateTransformations`**: Transformations between two or more images
-  MUST be stored in the attributes of a [`scene` object](#scene-md) in a [scene Zarr group](#scene-format).
-  In this case, the `input` and `output` values are objects
-  that refer to coordinate systems in the same zarr.json or in the metadata of multiscale image subgroups.
+  - One of `input` or `output` MUST reference the "intrinsic" coordinate system by `name` (`path` MAY be omitted or null).
+  - The other MUST reference a named coordinate system in the same multiscales group (by `name`, `path` MAY be omitted or null), or in a child [labels](#labels-md) group (by `name` and `path`).
```
btbest, line 468:
```suggestion
  - The other MUST reference a named coordinate system either within the same multiscale's `coordinateSystems` (by `name`, `path` MAY be omitted or null), or in a child [labels](#labels-md) group (by `name` and `path`).
```

Easy to forget, but a multiscales group can still contain multiple multiscales...

index.md
```diff
@@ -484,15 +461,17 @@ Depending on which, different constraints apply to the transformations, as descr
...
+  - When referencing a child labels group, the transformation MUST be [`identity`](#identity-md), [`scale`](#scale-md), or [`translation`](#translation-md).
```
btbest, line 469: Need to check again but aren't labels required to be the same shape as the multiscale; which means this basically always has to be identity? Allowing scale and translation isn't a problem of course, just presents more opportunities for confusion I suppose.

index.md
```diff
@@ -484,15 +461,17 @@ Depending on which, different constraints apply to the transformations, as descr
...
+
+- **Inside `scene > coordinateTransformations`**: Transformations between two or more images.
+  - Both `input` and `output` MUST contain `name`.
```
btbest:
```suggestion

- **Inside `scene > coordinateTransformations`**: Transformations between two or more multiscales.
  - Both `input` and `output` MUST specify a coordinate system `name`.
```

index.md
```diff
@@ -1351,13 +1323,12 @@ is the coordinate system that is referenced by all multiscale coordinate transfo
...
-  - The value of `input` MUST be the "intrinsic" coordinate system, referenced by `name`.
-    The `path` field of `input` SHOULD be omitted.
-  - The value of `output` can be a coordinate System in the same multiscales group (referenced by `name`).
-    In this case, the `path` field of `output` SHOULD be omitted.
-  - The value of `output` can be a coordinate system in a multiscales group in a child [labels](#labels-md) group (referenced by `path` and `name`).
-    In this case, the used transformation MUST be one of [`identity`](#identity-md), ['scale'](#scale-md) or ['translation'](#translation-md) transformations.
+  The following constraints apply:
+  - One of `input` or `output` MUST reference the "intrinsic" coordinate system by `name` (`path` MAY be omitted or null).
+  - The other of `input` or `output` MUST reference either:
+    - A named coordinate system in the same multiscales group (by `name`, `path` MAY be omitted or null), or
```
btbest, line 1372:
```suggestion
    - A named coordinate system in the same multiscale (by `name`, `path` MAY be omitted or null), or
```

btbest (approved):
Looks good as is, I just added some minor phrasing suggestions.

Remembering that any multiscales group can technically define multiple multiscales actually presents a semantic problem: the `name + path` referencing isn't ambiguous if the multiscales group at the `path` defines multiple multiscales whose intrinsic systems are identically named. I doubt this would be an issue in practice because nobody uses multiple entries in the multiscales array, but it's theoretically there.

I think to have version 0.6 being clean, we'd need the "The `multiscales` array MUST contain exactly one multiscale object` constraint that has been discussed before (but I think got delayed because it was bundled with flattening the multiscales-list to a single object, making it a breaking change for almost 0 gain).

index.md
jo-mueller: That's the part where we are entering the grey zone between collections and transforms. The relationship between labels and images is first and foremost semantic in nature, not spatial. Hence 0.6 will not force you to add spatial information to your label/image connection, but if you want to, you *can*. 

Hence, the shaping of the labels has never been written out in previous versions and 0.6 won't force this because only collections will express this relationship properly.

jo-mueller:
> Remembering that any multiscales group can technically define multiple multiscales actually presents a semantic problem: the name + path referencing isn't ambiguous if the multiscales group at the path defines multiple multiscales whose intrinsic systems are identically named. I doubt this would be an issue in practice because nobody uses multiple entries in the multiscales array, but it's theoretically there.

*Gulp*...there's an entire [RFC](https://ngff.openmicroscopy.org/rfc/6/index.html) around this (with a complicated history, I think ^^"). Essentially, it was decided to also defer to this to collections, because collections will arrange this bit of metadata *anyway*. As you say, so far there haven't really been any datasets in the wild that would make use of the array structure under the `multiscales` field so it's something I think we can live with in the meantime.

jo-mueller:
Follow-up: Since we are using the term `multiscale group` very extensively throughout the spec I'll avoid changing this terminology too much here, but this doesn't mean that it couldn't be changed in a follow-up PR (even in an 0.6.1 version, since this is really just phrasing, not semantics).

jo-mueller:
I agree that this statement would be the cleanliest way out of that issue:

> The multiscales array MUST contain exactly one multiscale object

But given the discussion around it, I think it's safer to just let it rest as it is for the moment and save the cleaning up for 1.0 😬

index.md
btbest: > The relationship between labels and images is first and foremost semantic in nature, not spatial

Subtle disagreement from the spec itself there:

> Some applications–notably image segmentation–produce a new image that is **in the same coordinate system as a corresponding multiscale image** (usually having the same dimensions and coordinate transformations).

Sounds very much spatial to me :)

Just to be clear, I'm not actually asking for anything to be changed here. Just indulging.

jo-mueller: Oh, I see what you mean. The loophole here is "usually"; There was a PR where thought of changing that (possibly #117) and require this transformation to be explicit. Ultimately we decided against it:

- If that transformation were to live in the parent multiscale image, then the addition of a child `labels` image would effectively invalidate the parent image because that transform would then be missing
- If the transform were to live in `/labels/segmentation_image/zarr.json`, its output would have to be `../../zarr.json`, which would *also* be problematic because moving the label image elsewhere would make it invalid and thus no longer self-describing.

The clean option would be to put the label image into a transformation graph and not use the `/labels/...` structure at all. *Or* wait for collections that will make this kind of linkage much more explicit from the point of view of an apex zarr.json in the hierarchy, which would be my preferred solution :)

btbest: I also don't see any point to requiring the metadata to enumerate all contents. At most I'd go to "SHOULD" level - just like the `labels` metadata only "SHOULD" reference all label multiscales available.

Issue #138, opened 2026-06-11
Remove "strict" jsonschemas and test cases


clbarnes:
For context, the "strict" cases treat any SHOULD in the spec as a MUST.

IMO this does not provide a lot of value: these conformance tests mainly target readers, and readers should be permissive about what they read. UIs of software working with OME-Zarr should encourage people towards writing SHOULD-conformant data but this doesn't really intersect with the use case of the conformance tests (or indeed the jsonschemas, which only apply level 1 validation anyway).

Meanwhile, they double the maintenance burden of the jsonschemas, which is not too bad, but also the conformance tests, which is a big deal, as we would eventually like to have a very comprehensive suite and every spec change can mean digging through every test case. If we can halve the effort with little loss of utility, that would be great.

Discussed with @kevinyamauchi @DragaDoncila @jo-mueller to generally positive feedback.

PR #140, opened 2026-06-11
Remove strict jsonchemas and conformance test cases 

clbarnes:
xref #138 .


will-moore:
Sounds fine to remove the strict schemas if they are more trouble than they're worth.

It seems that in some cases, e.g. well, multiscales and plate, we don't have any "non-strict" examples, so this PR removes the only examples we have.
Does that mean we don't have any non-strict tests for Wells / Plates?

Are the examples in this repo used ONLY for testing? Are they also useful for docs/ illustration examples too?

jo-mueller:
> Does that mean we don't have any non-strict tests for Wells / Plates? 

I think in these cases the strict/non-strict schemas are identical?

clbarnes:
Superceded by #176

PR #142, opened 2026-06-11
Change status from w3c/CG-FINAL to LS

### LINKED ISSUE #141: [Issue]: bikeshed builds failing
### Describe the issue

An index.bs of ours (https://github.com/ome/ngff-spec/blob/0.2/index.bs) is failing to render (https://app.readthedocs.org/projects/ngff/builds/33090983/)

With

 FATAL ERROR: Couldn't find an appropriate include file for the defaults inclusion, given group='None' and status='CG-FINAL'. `

Though it was just working fine in the past. 

bikeshed version is pinned, but builds are failing. 

there are some remote data assets that are independent of the pinned version; those are updated in the RTD builds

https://github.com/speced/bikeshed/commits/main/bikeshed/spec-data/readonly/boilerplate/doctypes.kdl was changed recently. that is likely the cause of the different build 

Probably we need to fix index.bs in all the spec version branches (0.1, 0.2, 0.3, 0.4, 0.5), but this may come to changes styling of the spec (so not so minor)

lubianat:
I think that closes #141 

It may change the layout also (I have not tested the build)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2,7 +2,7 @@
 Title: Next-generation file formats (NGFF)
 Shortname: ome-ngff
 Level: 1
-Status: w3c/CG-FINAL
+Status: LS
 TR: https://ngff.openmicroscopy.org/0.2/
 URL: https://ngff.openmicroscopy.org/0.2/
 Repository: https://github.com/ome/ngff


lubianat:
not the fix

PR #144, opened 2026-06-11
docs: add recommendation for coordinate system

jo-mueller:
This PR adds a hint under the scene metadata section to establish some guidelines on which coordinate systems viewers should choose, depending on whether 
- coordinate systems are present in the scene metadata (if so, viewers SHOULD default to the first one)
- coordinate systems are only present in the multiscales metadata - in this case, viewers MAY pick any (i.e. the first)

### DIFF
--- a/index.md
+++ b/index.md
@@ -1978,6 +1978,14 @@ A transformation from instrument3 to instrument1 can be obtained
 by composing the two transformations above.
 :::
 
+```{hint} Which coordinate system to display?
+Writers of [`scene`](#scene-md) metadata can signal to viewers which coordinate system use for visualization.
+If a coordinate system is defined under the `coordinateSystems` field of the `scene` metadata, then viewer SHOULD treat the first coordinate system in that list as the default coordinate system for display.
+If no coordinate system is defined therein, but only in the respective linked multiscale groups,
+the first coordinate system in the list of parsed coordinate systems MAY be used as the default coordinate system for display.
+
+```
+
 ## Specification naming style
 (naming-style)=
 
--- a/index.md
+++ b/index.md
@@ -1979,10 +1979,16 @@ by composing the two transformations above.
 :::
 
 ```{hint} Which coordinate system to display?
-Writers of [`scene`](#scene-md) metadata can signal to viewers which coordinate system use for visualization.
-If a coordinate system is defined under the `coordinateSystems` field of the `scene` metadata, then viewer SHOULD treat the first coordinate system in that list as the default coordinate system for display.
+When reading [`scene`](#scene-md) metadata,
+some applications may want to choose a coordinate system as a default reference,
+e.g. for immediate display.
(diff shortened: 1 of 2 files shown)


jo-mueller:
cc @kevinyamauchi @tischi @aafkegros @btbest @bogovicj @clbarnes

index.md
```diff
@@ -1978,6 +1978,14 @@ A transformation from instrument3 to instrument1 can be obtained
 by composing the two transformations above.
 :::
 
+```{hint} Which coordinate system to display?
+Writers of [`scene`](#scene-md) metadata can signal to viewers which coordinate system use for visualization.
+If a coordinate system is defined under the `coordinateSystems` field of the `scene` metadata, then viewer SHOULD treat the first coordinate system in that list as the default coordinate system for display.
+If no coordinate system is defined therein, but only in the respective linked multiscale groups,
+the first coordinate system in the list of parsed coordinate systems MAY be used as the default coordinate system for display.
```
LucaMarconato: I would make it more explicit, such as "the input coordinate system of the first transformation". But otherwise looks good!

clbarnes:
I have a soft opinion that opening any particular coordinate system is not necessarily the right behaviour for a viewer. If the wrong coordinate system is opened, changing it in the GUI is likely to be more annoying than asking a user what they want up front. Opening an undesired coordinate system could be costly if it by default opens many co-registered images and/or if some images use vector field transformations.

The language could be softened to something like

> Viewer applications opening a Scene for the first time MAY allow users to select which coordinate system and which images they wish to view.
> If they instead choose to show a default coordinate system, it SHOULD be the first.
> It is RECOMMENDED that writers of `scene` metadata order their coordinate systems to support this pattern.

What do HDF5-based viewers do?

btbest:
I'm also in favour of softer phrasing (not that should/may really make much of a practical difference...)

I'm happy with @clbarnes  phrasing, but here's another angle I could suggest:

> When reading `scene` metadata, some applications may want to choose a coordinate system as a default reference, e.g. for immediate display. If they do so, it is RECOMMENDED that the `scene`'s first `coordinateSystems` entry is used as the default, if present. Likewise, when writing `scene` metadata, if there is a coordinate system that acts as a common reference for other coordinate systems within the `scene` or its child multiscales, it is RECOMMENDED that this coordinate system be stored as the first `coordinateSystems` entry in the `scene`.

jo-mueller:
I like the suggestion of @btbest! Will change :)

clbarnes (approved):

jo-mueller:
@kevinyamauchi @tischi @aafkegros @btbest @bogovicj last call for opinions - will merge in the coming days :)

aafkegros:
I think this is good for now, I feel part of this may become deprecated when a more full-fledged 'viewing' RFC comes in but is good to have as a standalone recommendation.

PR #145, opened 2026-06-12
Displacements are multiscales

### LINKED ISSUE #520: RFC-5: relocate coordinates/displacements index transformations

jo-mueller:
Fixes https://github.com/ome/ngff/issues/520

This changes how displacement fields and coordinate arrays are expected to be stored in displacement/coordinate transforms, as discussed in the 06/2026 Hackathon at EMBL.

The gist: The arrays are now simply multiscale groups and the displacement/coordinate dimension should go into the channel dimension of the data.

- updated the spec text
- updated the examples (removed the 1d examples because 1d is not allowed anyway)
- no change to the schema
- updated version history

@bogovicj @clbarnes @melonora @lucamarconato @thewtex @tischi @kevinyamauchi @jni @dragadoncila @perlman
...who am I forgetting?


jo-mueller:
Maybe should add a recommendation/statement about which resolution level of the displacement field implementations should use?

*If multiple resolution levels are provided, implementations MAY choose any resolution level to compute the output.*

Also, do we need to add constraints/hints on possible pathologies that can arise when downsampling vector fields/coordinate lists?

thewtex (approved):
🥇 😍 🚀

thewtex:
> Maybe should add a recommendation/statement about which resolution level of the displacement field implementations should use?
> 
> _If multiple resolution levels are provided, implementations MAY choose any resolution level to compute the output._

👍 

And possibly SHOULD use the finest resolution if the result is used beyond quick visualization purposes.

> Also, do we need to add constraints/hints on possible pathologies that can arise when downsampling vector fields/coordinate lists?

We could also recommend that anti-alias filtering SHOULD be applied when downsampling, similar to an intensity image multiscale.

index.md
```diff
@@ -1088,36 +1088,17 @@ implementations of the specified interpolation methods may still differ in their
...
-  - **Dimensionality**: If the input coordinate system has `N` axes, the array at location `path` MUST have `N+1` dimensions.
+The multiscale group at `path` MUST satisfy:
+  - **Dimensionality**: If the input coordinate system has `N` axes, the multiscale image at location `path` MUST have `N+1` dimensions.
   - **Vector dimension length**: 
     - For `coordinates` transformations, the length of the array along the `coordinate` dimension (last axis) MUST equal `M`,
       the number of axes in the output coordinate system.
+      The coordinate dimension SHOULD be stored as the channel axis (`"name": "c"`) of the multiscale image group and have `"type": "coordinate"`.
     - For `displacements` transformations, the length of the array along the `displacement` dimension (last axis) MUST equal `N`,
       the number of axes in the input (and output) coordinate system. `displacements` require `M=N`.
+      The displacement dimension SHOULD be stored as the channel axis (`"name": "c"`) of the multiscale image group and have `"type": "displacement"`.
```
d-v-b: I wouldn't overload the concept of multichannel images for this purpose. A displacement isn't like a channel.

jo-mueller: It's not, but we figured that until RFC3 is pushed over the line, it would be easier this way for several reasons:
- It's not a MUST - you can also call the used axis `d`
- The type `"displacement"` clearly indicates what it's used for
- RFC3 will open/improve the usability of more custom axes.

It's not ideal, but solving this "properly" would turn 0.6 into containing RFC3, too.

d-v-b: I might be missing something, but I thought the concept of a "channel" axis was defined entirely w.r.t to its `type` field, not its name. So why is the name mentioned here? Seems like you can leave the name uncontrolled and just require the type, and then you don't need any references to "the channel axis"

d-v-b: also, you should probably remove the arbitrary axis restrictions as soon as possible. Otherwise 5 dimensional data can't have displacement / coordinate transforms, because the resulting transform array would be 6 dimensional, and therefore invalid. You can of course work around this with dropaxis and whatnot, but given that the dimensionality restriction is completely arbitrary, I would fix that first.

jo-mueller: > Seems like you can leave the name uncontrolled and just require the type, and then you don't need any references to "the channel axis"

That's a good suggestion I could see myself getting behind if it's a blocker for others, too. People could still name the axis `c` (mostly because that's what implementations expose) and as long as `type: displacement` is tied down that'd be good enough for me.

> also, you should probably remove the arbitrary axis restrictions as soon as possible.

RFC3 is next on the menu, but that's for post-0.6.

clbarnes: From the existing 0.6 draft, emphasis mine:

> axes entries MUST be ordered by type where the time axis must come first (if present), followed by the channel * * * or custom * * * axis (if present) and the axes of type space.

From v0.5:

> The order of the entries MUST correspond to the order of dimensions of the zarr arrays. In addition, the entries MUST be ordered by “type” where the “time” axis must come first (if present), followed by the “channel” * * * or custom * * * axis (if present) and the axes of type “space”.

That is, we don't currently intrinsically have a channel axis. The slot which the channel axis must occupy is also allowed to be a null or custom -typed axis. It's probably better not to normatively refer to it as a channel axis, but say that it MUST be after the time axis, MUST be before the space axes, and MUST not coincide with a channel or null or custom axis.

d-v-b: > That is, we don't currently intrinsically have a channel axis. 

yeah this is my main thing. The first reference to "_the_ channel axis" should probably not be here :)

jo-mueller: Better in current state? Thanks @clbarnes for the text bricks 👍

d-v-b: it's improved! Also, is it an explicit goal to prevent folks from storing their coordinates as an array of N-dimensional points (i.e., the zarr data type says "each element is a point"), instead of an N+1 dimensional array?

clbarnes: IMO, yes. We talked about being conservative about depending on zarr extensions and may in future recommend that users use core rather than extension codecs and data types where possible. Here we know the output should be an array of numbers, and arrays of numbers are universally supported by zarr implementations; I don't see a compelling reason to depend on support for structs or fixed-size lists. I would also recommend picking one representation rather than putting more ORs in the spec.

jo-mueller:
@bogovicj @clbarnes @melonora @LucaMarconato @thewtex @tischi @kevinyamauchi @jni @DragaDoncila @perlman @d-v-b last call - any more feedback on this?

LucaMarconato:
One comment not in the diff of this PR:

> - If `nearest` interpolation is used for a `coordinates` transformation,
>   the transformed image collapses into a single point at the nearest coordinate in the coordinate field.
> - If `nearest` interpolation is used for a `displacements` transformation,
>   the transformed image is piecewise constant with discontinuities at the boundaries between nearest neighbor regions.

I haven't worked with this yet so I may be missing something, but aren't the two representations (coordinate and displacements) equivalent? Why do we have 2 different "if"? Shouldn't be the implication the same?

EDIT: ah I see below that "`displacements` require `M=N`." so they are not equivalent and now I got what the 2 "if" mean: with displacement the shift is constant for all the neighbors of a point in the domain grid, while for coordinates the target image is constant for each point of the domain grid.

So the text it's correct, but it can be confusing if someone is new with the concept because the definition of how displacement and coordinates operates as a function is missing. The specs start right away in showing how to represent them, but not how they would behave as functions.

index.md
```diff
@@ -1089,36 +1089,16 @@ implementations of the specified interpolation methods may still differ in their
...
-  - **Vector dimension length**:
+The multiscale group at `path` MUST satisfy:
+  - **Dimensionality**: If the input coordinate system has `N` axes, the multiscale image at location `path` MUST have `N+1` dimensions.
+  - **Vector dimension length**: 
     - For `coordinates` transformations, the length of the array along the `coordinate` dimension (last axis) MUST equal `M`,
       the number of axes in the output coordinate system.
     - For `displacements` transformations, the length of the array along the `displacement` dimension (last axis) MUST equal `N`,
       the number of axes in the input (and output) coordinate system. `displacements` require `M=N`.
+    - The axis ordering of the displacements/coordinates array MUST satisfy the constraints of [multiscale image groups](#multiscales-md). In that array, the array dimension corresponding to the coordinate/displacement axis MUST be inserted after a time axis (if present) and before the spatial axes and MUST not coincide with a channel or custom axis in the input coordinate system. The axis MUST specify ``type: "displacement"`` or ``type: "coordinate"`` in the multiscale image's [axes metadata](#axes-md).
```
LucaMarconato, line 1101: Worth adding a paragraph that when RFC-3 lands we need to remove this restriction. Or we need to add a reminder for this directly in the RFC-3 text.

index.md
```diff
@@ -1089,36 +1089,16 @@ implementations of the specified interpolation methods may still differ in their
...
   - **Vector component mapping**: The `i`th value of the array along the `coordinate` or `displacement` axis refers to the `i`th output axis.
```
LucaMarconato, line 1102: Comment to 

>   - **Vector component mapping**: The `i`th value of the array along the `coordinate` or `displacement` axis refers to the `i`th output axis.

For displacement the `i`th component of the array corresponds to `i`th component of the input axis, but input and output are assumed to coincide (in dimension, since we say that `N`=`M` and also in order). I would make these conditions more explicit in the text.

index.md
jo-mueller: So far the spec doesn't contain any forward references to how things may or may not change in the future. But this is definitely something on the plate of RFC3 - generally, the interaction between RFC5 and 3 will be interesting, but I'd argue that this goes for any transform when RFC3 lands.

index.md
jo-mueller: > For displacement the `i`th component of the array corresponds to `i`th component of the input axis

Wouldn't it actually be the `i+1`th axis in the displacements array, since the latter has one more dimension than the input array, that prepends the spatial axes, i.e, a `zyx` image needs a `czyx` displacements field unless it's wrapped in a `byDimension`

index.md
```diff
@@ -1133,168 +1113,30 @@ Applying the transformation to a point `x` in the input coordinate system amount
    - a displacement to add to the input point `x` (`displacements`).
 ```
```
LucaMarconato, line 1117: I don't get it, I don't find this clear. The text is probably correct for someone that knows already how these things work, but I don't think it's unambiguous/clear enough.

1. We said that we don't want to compute the inverse, so I guess the displacement is actually to be interpeted as a inverse transformation?
2. What is a? f(x)? f-1(x)? I don't understand. 
3. In "interpolate the vector field to obtain a vector for the input point" are we talking about the transformation or the inverse transformation?
4. In the vector that we obtain above, does the vector belong to the original transformation or the inverse?

I have 2 suggestions:
1. make a drawing for the displacement case and one for the coordinate case. 
    1.a. Make clear in which direction is the transformation, and if the inverse transformation is the initial transformation being interpreted as directed the other way, or is a derived approximation.
    1.b. Define what is a and show it in the figure. How is a different from f(x)?
2. consider discussing the displacement and coordinate cases distinctly, or if not, at least making it clearer when we are in one case or the other. 
3. explain the direction of the vector (x -> a or a -> x?)show in the figure what is the vector obtained by "interpolate the vector field to obtain a vector for the input point"

index.md
```diff
@@ -1133,168 +1113,30 @@ Applying the transformation to a point `x` in the input coordinate system amount
...
-  }
-]
+```{literalinclude} examples/transformations/displacements/multiscales.json
+:language: json
 ```
 
-The metadata at location `displacementField` should have a coordinate system such as:
+This multiscales (containing the vector field) MUST have three dimensions,
+i.e., `x` and `y` as mandated by the input coordinate system and an additional channel dimension `d` to hold the x- and y-displacements.
+The metadata for the multiscale group at location `coordinateTransformations/displacementField` would look as follows;
```
LucaMarconato: truncated sentence/missing example?

index.md
```diff
@@ -1133,168 +1113,30 @@ Applying the transformation to a point `x` in the input coordinate system amount
...
 Indexing into this array using c-order, for spatial positions `y` and `x`, the y- and x-displacements would be given by:
 
 ```
-y_displacement = displacementField[y][x][0]
-x_displacement = displacementField[y][x][1]
+y_displacement = displacementField[0][y][x]
+x_displacement = displacementField[1][y][x]
 ```
 
 I.e. the y-displacement is first, because the y-axis is the first element of the input and output coordinate systems.
```
LucaMarconato, line 1144: This refers to the old example where the access was [y][x][0], now it's [0][y][x]. So either this sentence or the pseudocode is incorrect.

index.md
LucaMarconato: This sentence

> - For `displacements` transformations, the length of the array along the `displacement` dimension (last axis) MUST equal `N`,
>       the number of axes in the input (and output) coordinate system. `displacements` require `M=N`.

is why I said M=N (and `i`th mapping to `i`th). I don't think the `c` is being considered here. My understanding is the following:
- the displacements function maps: zyx -> zyx. To define it we need the shifts, which are described by 3 numbers. So displacements in this case require a 4d tensor of shape (length_z, length_y, length_x, 3) (for each zyx position we need 3 numbers).
- the coordinate function maps zyx -> kczyx (basically any output space). To define it we need the target coordinates. So coordinates in this case require a 4d tensor of shape (length_z, length_y, length_x, 5) (for each zyx position we need 5 numbers).

LucaMarconato: If the above is correct, my initial comments stands valid.

LucaMarconato:
Thanks @jo-mueller for opening this! My comments are most about the text that is not touched by this PR, but that I reviewed together with the changes. I think the changes are mostly ok (and with the new usage of multiscales the text is shorter and clearer).

index.md
jo-mueller: I think you got it right by 90%, it also took me quite a while to wrap my head around it.

1. Yes, transformations are *always* expressed in the forward direction, I.e., a displacements vector points from a location in the input space to a location in the output space.
2. The displacement vector field is *sampled regularly* on the space of the input image, but the sampling density can vary. I.e., for a 16x16 image, you may have a `2x16x16` vector field (one vector for every pixel), or something like `2x4x4` (one vector every four pixels).
  To make this explicit, the displacements field needs to declare this sampling density. The easiest thing to do is to use a scale transform, that maps the array coordinates for the displacement field into physical coordinates.
3. This scale transform is *forward* (displacement field array coordinates -> physical space). To find the corresponding vector array coordinate for a given physical location, one needs to invert the scale transform. 
4. The displacement vector field is then interpolated at the derived vector array coordinate (which may or may not coincide with an integer value).

I'll try myself at improving the language here 👍

index.md
LucaMarconato: Makes sense, we can hit "Resolve" on this.

index.md
jo-mueller: Jup! 1eb130456c87cb7ece59c9aa3400cb032426c9ec

index.md
jo-mueller: I think it's actually correct. The axes order is always `cyx`, which means that that's the order by which the displacements field is indexed. The vector that is stored along the `c` axis follows the same conventions, so that `y=vector[0]` and `x = vector[1]`. I updated in 12497b73f115de8c9bd2c2d0b9ce5dbd2f206613 to reflect this better!

jo-mueller:
@LucaMarconato thanks for the review! I expanded the example below the hint box a bit to make the procedure more clear on how the retrieved, interpolated vector is obtained from a given input point `x` in physical coordinates.

I'm considering the `hint` box entirely since the example is maybe more informative anyway. Let me know if that clears things up!

index.md
```diff
@@ -1102,11 +1102,14 @@ The multiscale group at `path` MUST satisfy:
...
-  to map the input point `x` into the corresponding array coordinate `a`.
+Applying the transformation to a point `x` in physical coordinates in the input coordinate system amounts to following the following steps:
+1. The displacement vector field contains a mapping from its array coordinate system into the coordinate system,
+  which coincides with the input coordinate system of the transformation.
+  Hence, the coordinates of the input point `x` need to be mapped into the array coordinate system of the vector field to look up the corresponding vector in the array.
+  The inverse of the transformation found in the vector field's metadata under `coordinateTransformations`
+  is used to map the input point `x` into the corresponding array coordinate `x_a`.
 2. Look up the vector in the array corresponding to that point's coordinates in the array's coordinate system.
-3. If the point (`a`) does not correspond to a discrete point in the `coordinate` or `displacement` array,
+3. If the point (`x_a`) does not correspond to a discrete point in the `coordinate` or `displacement` array,
```
LucaMarconato, line 1112: I found it clear now, thanks!

index.md
```diff
@@ -1123,24 +1126,37 @@ The transformation references a displacement vector field located at `coordinate
 ```
 
 This multiscales (containing the vector field) MUST have three dimensions,
-i.e., `x` and `y` as mandated by the input coordinate system and an additional channel dimension `d` to hold the x- and y-displacements.
-The metadata for the multiscale group at location `coordinateTransformations/displacementField` would look as follows;
-The channel dimension corresponds to an axis with `type : displacement`,
-the other two dimensions MUST be axes that are identical to the axes of the input coordinate system (`"name": "physical"`) to which the transformation is applied.
+i.e., `x` and `y` as mandated by the input coordinate system and an additional dimension `c` to hold the x- and y-displacements.
```
LucaMarconato, line 1129: Do we really want to call this `c`? It can be confusing with the `c` of an image, in case we are applying a displacement `c` to an image instead of a label. I think `d` (which was there before the edit), was fine.

index.md
LucaMarconato: True, and the new text makes it clearer.

index.md
d-v-b: why does the name matter at all? there's a `type` field that implementations can use

jo-mueller: Agree with @d-v-b. The type field has the authoritative information, and the usage of `c` is ambiguous already (channel? custom? coordinates? - all are terms appearing in the spec) so that overloading it with a particular meaning in this context will only lead to confusion elsewhere.

index.md
jo-mueller: Great, thanks for the feedback! 🎈

index.md
melonora: While the spec currently does not directly contain forward references, I refer here to RFC-4 in which the RFC is written as linkml. Here it is written in such a way that it would be easy to add this to the spec: https://github.com/ome/ngff/blob/1ea8323282ec0f643b601e72a21553e54e53e7c7/rfc/4/orientation.yml#L27-L42

Just add an extra axis here under `any_of`. I also told @jni that the way it is written here, pretty much directly allows for more axes for thee.

jo-mueller: schemas as linkML would be very desirable :)

index.md
jo-mueller: 👍 for displacements except for that the displacements tensor should have `(3, length_z, length_y, length_x)`. The crucial point is, that to apply the vector, you need a base point plus the displacement. The base point is given by the index in the displacements array. 

If you have a base point with array coordinates `[i, j k]`, these coordinates can be mapped to physical space using the multiscale transformations (i.e. `"scale": [0.5, 0.5, 0.5]`). The corresponding direction component of the vector is `dfield_array[:, i, j, k]`

The same goes for the coordinates and it also took me a while to wrap my head around this: The coordinate array is not en enumeration of pairs of points, but the *index of the array is the input point*. Say, for a given input point with coordinates `[i, j, k]` (which can again be mapped into physical space using the multiscale transforms), you'd find the corresponding output point by indexing the coordinates array at that location:

`output_point_coords = coord_array[:, i, j, k]`

Hence the restriction on the array dimensions! The size of the first axis of the displacement/coordinates array indicates the dimensionality of the output space, which may be different from the input space!

@bogovicj please give me a 👍 or 👎 if I'm explaining this correctly!

melonora: In general, would it help that we maybe come up with a schematic figure to illustrate these concepts for externals? I think this might say more than a 1000 words to newcomers in this case.

index.md
melonora: Agree with @d-v-b as well and this also fits more with RFC3 in mind. @thewtex I will create a PR to RFC 4 because I think we defined `name` still as a defining slot which might be the case now, but not if we have RFC3 in mind.

joshmoore (approved):

version_history.md
```diff
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## {0.6rc0} - 2026-06-12
```
jo-mueller:
```suggestion
## 0.6 - 2026-06-12
```

index.md
bogovicj: Yes, @jo-mueller 's  description is correct! 

> The size of the first axis of the displacement/coordinates array indicates the dimensionality of the output space, which may be different from the input space!

Exactly. Small clarification: the size of the output space can differ from the input space for `coordinates` but they have to be the same for `displacements`. Reason being is that `displacement` store a vector that is added to the input point, so we constrain things so we add vectors of the same length.

jo-mueller: Thanks @bogovicj for the feedback! Will resolve, we can always add more descriptive text or figures later (even after release) without changing the actual spec.

clbarnes (approved):

PR #146, opened 2026-06-15
spec: Hint on expectations on multiscale transforms

### LINKED ISSUE #494: multiscales.datasets coordinate transformations: example inconsistency and spec clarification
I'm opening this issue to write down two points that came to me regarding transformations and resolution levels in v0.6:

1) An inconsistency in two examples
2) Ask whether the spec could benefit from a quick clarification

Warning: the text here ended up being rather long but I hope/promise the point made is relatively concise 🤗🙏
## Context

In the v0.6 spec, the multiscales metadata contains coordinateTransformations for the objects listed in multiscales.datasets. These transformations make sure all resolution levels are consistently represented in the "intrinsic coordinate system". 

For each resolution, the ratio between the parameters of the scale transformation of the current resolution and the first resolution represents the scaling factor between the resolutions. A translation transformation helps to align potential offsets of the different resolutions.

In practice, these offsets depend on the downsampling method that's been used to create the arrays representing the different resolutions. For many downsampling methods (most in practice?), e.g. when using classical binning in which the binning window is aligned with the image origin (think of the first three pixels being summarised in a single pixel), a translation transformation needs to account for the fact that the first pixel of the current resolution no longer maps to the same coordinate as the first pixel of the first resolution in the "intrinsic coordinate system" (due to the [coordinate convention](https://ngff-spec.readthedocs.io/en/latest/#coordinate-convention)).

## multiscales examples

So most downsampling methods will require a (non-zero) translation transformation to be set on all resolutions but the first one. See also the discussion [here](https://github.com/ome/ome-zarr-py/issues/403)](https://github.com/ome/ome-zarr-py/issues/403).

The example [multiscales_strict/multiscales_example.json](https://github.com/ome/ngff-spec/blob/main/examples/multiscales_strict/multiscales_example.json) represents a multiscale image with three resolution levels. The coordinate transformations associated to each resolution level only contain a scale transformation each. This is not by itself inconsistent, because a downsampling method could have been used which actually doesn't require a translation transform to align the resolutions. However, the example explicitly states in the `multiscales.metadata` field that `skimage.transform.pyramid_gaussian` has been used for downsampling. This downsampling method requires a translation transformation as described [here](https://github.com/ome/ome-zarr-py/issues/403).

The other multiscales example with more than one resolution [multiscales_example_relative.json](https://github.com/ome/ngff-spec/blob/main/examples/multiscales_strict/multiscales_example_relative.json) also doesn't include translation transformations. However it doesn't explicitly mention a downsampling method, so the example is not technically inconsistent, but perhaps slightly misleading.

Next to the inconsistency, I think it's a bit confusing that the two examples don't represent a realistic use case and that an implementer who looks at them needs to find out by themselves that a translation transformation in practice is probably required. **Therefore: Should we add consistent translation transformations to the multiscales examples mentioned above?** Some explanatory comments could also be nice. There are more examples in the tests, but these are probably less confusing.

## Clarifications in the spec text?

Regarding this, what I've found in the spec text is the coordinate convention, explanations of the intrinsic coordinate system and the following paragraph:

>The transformation MUST be one of the following:
> - A single scale or identity transformation
> - A sequence transformation containing one scale and one translation transformation.

> In these cases, the scale transformation specifies the pixel size in physical units or time duration. If scaling information is not available or applicable for one of the axes, the value MUST express the scaling factor between the current resolution and the first resolution for the given axis, defaulting to 1.0 if there is no downsampling along the axis. This is strongly recommended so that the the “intrinsic” coordinate system of the image avoids more complex transformations.

While this is not a strictly technical interpretation, the fact that translation transformations are optional would suggest to me that these are useful for some advances use cases, but not generally needed. **Therefore, would it make sense to clarify this aspect in the spec text?**

One suggestion would be adding something like the following to the the paragraph above:

"Consistent positional alignment across resolutions in the intrinsic coordinate system must be ensured using translation transforms if required. E.g. for the most common downsampling methods such as classical binning, a translation of (pixel-size-at-resolution-N - pixel-size-at-resolution-0) / 2 is required for alignment."

jo-mueller:
Fixes https://github.com/ome/ngff/issues/494
Relevant for https://github.com/ome/ngff/issues/89

@thewtex @m-albert @jni @tischi @d-v-b some more language on the expectations around multiscale transformations as proposed by @m-albert over at https://github.com/ome/ngff/issues/494 (sorry for the long radio-silence about the issue 📻 )

### DIFF
--- a/index.md
+++ b/index.md
@@ -1476,8 +1476,8 @@ is the coordinate system that is referenced by all multiscale coordinate transfo
 
   ```{note}
   These transformations are expected to consistently align the datasets across all scales in the intrinsic coordinate system.
-  E.g. for the most common downsampling methods such as classical binning,
-  including translations of (pixel-size-at-resolution-N - pixel-size-at-resolution-0) / 2 for each axis is required for correct multi-scale alignment.
+  For the most common downsampling methods such as classical binning,
+  translations should be (pixel-size-at-resolution-N - pixel-size-at-resolution-0) / 2 for correct multi-scale alignment.
   ```
 
   In these cases, the scale transformation specifies the pixel size in physical units or time duration.
--- a/index.md
+++ b/index.md
@@ -1474,6 +1474,12 @@ is the coordinate system that is referenced by all multiscale coordinate transfo
   * A single scale or identity transformation
   * A sequence transformation containing one scale and one translation transformation.
 
+  ```{note}
+  These transformations are expected to consistently align the datasets across all scales in the intrinsic coordinate system.
+  E.g. for the most common downsampling methods such as classical binning,
+  including translations of (pixel-size-at-resolution-N - pixel-size-at-resolution-0) / 2 for each axis is required for correct multi-scale alignment.
+  ```
+
   In these cases, the scale transformation specifies the pixel size in physical units or time duration.
   If scaling information is not available or applicable for one of the axes,
   the value MUST express the scaling factor between the current resolution


index.md
```diff
@@ -1474,6 +1474,12 @@ is the coordinate system that is referenced by all multiscale coordinate transfo
   * A single scale or identity transformation
   * A sequence transformation containing one scale and one translation transformation.
 
+  ```{note}
+  These transformations are expected to consistently align the datasets across all scales in the intrinsic coordinate system.
+  E.g. for the most common downsampling methods such as classical binning,
```
thewtex: "E.g. for" -> "For"

index.md
```diff
@@ -1474,6 +1474,12 @@ is the coordinate system that is referenced by all multiscale coordinate transfo
...
+  including translations of (pixel-size-at-resolution-N - pixel-size-at-resolution-0) / 2 for each axis is required for correct multi-scale alignment.
```
thewtex: "including translations of" -> "translations SHOULD be "
"is required for correct multi-scale alignment" -> " for the correct multi-scale alignment"

thewtex (commented):
@jo-mueller thanks for getting this in!

A few wording suggestions inline.

index.md
jo-mueller: 7a405c76f9fbf4fa14cb2df71bc56e601261d261

index.md
jo-mueller: 7a405c76f9fbf4fa14cb2df71bc56e601261d261

thewtex (approved):
🥇

m-albert (approved):
Thanks @jo-mueller and @thewtex 🚀

jo-mueller:
Thanks @thewtex @m-albert for the feedback. Would you mind hitting the `approve` button here? Makes it easier for us to go through all of these open PRs a bit quicker :)

Edit: nvm, just saw you already did. Thanks!!

jo-mueller:
@joshmoore if @clbarnes approves I think this is one we can also merge async.

clbarnes (approved):

PR #154, opened 2026-06-23
add Assa Diabira as author

assadiab:
Following my RFC-9 comment (ome/ngff#544) and @jo-mueller's suggestion.

Added my entry to `authors.yml` and `CITATION.cff`. I did not regenerate `CITATION.cff`, as it is currently out of sync with `authors.yml` and would include other pending author updates.


joshmoore:
Welcome, @assadiab! 🎉 🙌🏽

PR #158, opened 2026-06-26
spec: tighten rules for metadata on displacements

jo-mueller:
@clbarnes @bogovicj @kevinyamauchi @jni @d-v-b @thewtex 

Some additional wording around requirements for the multiscale image that stores the displacement/coordinate vector field to signal explicitly from which axis exactly to retrieve the coordinate/displacement vector.


thewtex (approved):
👍

jo-mueller:
@clbarnes @bogovicj @kevinyamauchi @jni @dragadoncila @will-moore last chance on opinions - formal "approve" will help for decision history tracking 👍

index.md
```diff
@@ -1056,6 +1056,9 @@ The array containing the coordinates or displacements MUST:
   to the corresponding coordinates in the input coordinate system via a coordinate transformation (see details below).
 - have one dimension corresponding to every axis of the input coordinate system
 - have one additional dimension to hold components of the vector (either coordinates or displacements)
+- specify `"type": "displacement"` or `"type": "coordinate"` depending on the type of transformation and
+ `"discrete": true` in the multiscale image's [axes metadata](#axes-md)
```
d-v-b, line 1039: how is this field interpreted w.r.t. the recommended interpolation scheme for this displacement / coordinate field?

jo-mueller: The `discrete` flag communicates that interpolation along this axis is weird, essentially. RFC3 will have to generalize this further, but essentially this is a good way of telling an implementation that this axis behaves like a channel without referring to it as "channel".

Edit: More in the context of coords/displacements, the axis "type" signals in which axis the vectors/coordinates are stored without having to specify a [projectAxis ](https://github.com/ome/ngff-spec/pull/130) transform that somehow is and isn't part of the transform graph.

d-v-b: that makes sense. as with channels, it's a bit unfortunate that there's no way to use the axis metadata to indicate what indices along the coordinate / displacement axis mean (are we indexing x,y,z vectors or z,y,x?). Using a struct dtype would be a bit more elegant here. This is something an RFC should fix.

d-v-b: second question: are readers expected to error if discrete was not set to true, or if type was not set to displacement or coordinate?

jo-mueller: Yeah we also figured that in this context that's the next-best thing to make explicit how axes behave under transforms. To be fair, the axes metadata *kind of* provides this information but even there, names like czyx are "just" names without any further meaning. Only RFC4 will add more semantics there.

d-v-b: makes sense! my small worry was that _requiring_ readers to error here could make valid data unreadable if the writer wrote the coordinate / vector field but accidentally omitted the required axis annotation. but if all the writers are correct, this won't happen.

d-v-b (approved):

index.md
jo-mueller: As per the second question, I'd say: Be kind on read and strict on write. The spec communicates clearly here that if type is displacement, then discrete MUST also be true and I think this can be  expected from writers. But that's more of a matter of philosophy of a reader rather than something that the spec can (should?) enforce.

I'm not sure whether this can be expressed by json schemas, though 🤔

DragaDoncila (approved):
All good from me @jo-mueller!

jo-mueller:
@joshmoore also good to go 👍

index.md
d-v-b: > But that's more of a matter of philosophy of a reader rather than something that the spec can (should?) enforce.

This is actually something the spec can and probably should weigh in on. The spec can define the shape of "best effort" input parsing when inputs are truly parsable, while also defining conditions where readers MUST error. In this case, I would say that readers MUST error if the coordinate / displacement field has the wrong dimensionality, but I'm not sure I would require readers to error if the "discrete" field wasn't set correctly.

PR #161, opened 2026-06-29
examples: add `type` to axes metadata in examples

### LINKED ISSUE #160: [Issue]: inconsistency between wording of `multiscales.coordinateSystems` spec, and 2 example
### Describe the issue

Line ~255 of index, in the `"axes" metadata` section, says `axis.type` is technically optional

https://github.com/ome/ngff-spec/blob/36f74596b935606bfa53464ec18e808bce4b0531/index.md?plain=1#L255-L257

... while the `multiscales.coordinateSystems` section further restricts what `axes` must look like when they are inside multiscales metadata.  Specifically, that "`axes` MUST contain 2 or 3 entries of `type:space`":

https://github.com/ome/ngff-spec/blob/36f74596b935606bfa53464ec18e808bce4b0531/index.md?plain=1#L1280-L1285

So far, that's fine.  however, the `color_properties.json` example shows an `ome.multiscales.coordinateSystems` that *lacks* axes of `type:space`:

https://github.com/ome/ngff-spec/blob/36f74596b935606bfa53464ec18e808bce4b0531/examples/label_strict/colors_properties.json#L5-L17

... so I went to read the labels section for any obvious "except when it appears in labels", but I don't see anything obvious.  It does say "This image group MUST implement the multiscales specification" ... in which case the example seems to violate it.

----

**Question:**: when validating the `attributes.ome.multiscales[].coordinateSystems[]`, _should_ we be able to assert that there are that "there are 2 or 3 entries of `type:space`"?  Or is this an exception if that `ome.multiscales` happens to be a `labels`?

jo-mueller:
Fixes https://github.com/ome/ngff-spec/issues/160

### DIFF
--- a/examples/label_strict/colors_properties.json
+++ b/examples/label_strict/colors_properties.json
@@ -10,9 +10,9 @@
             {
               "name": "physical",
               "axes": [
-                {"name": "z", "unit": "micrometer"},
-                {"name": "y", "unit": "micrometer"},
-                {"name": "x", "unit": "micrometer"}
+                {"name": "z", "type": "space", "unit": "micrometer"},
+                {"name": "y", "type": "space", "unit": "micrometer"},
+                {"name": "x", "type": "space", "unit": "micrometer"}
               ]
             }
           ],
--- a/examples/multiscales_strict/multiscales_example_relative.json
+++ b/examples/multiscales_strict/multiscales_example_relative.json
@@ -6,13 +6,13 @@
       "coordinateSystems" : [
         {
           "name" : "world",
-          "axes": {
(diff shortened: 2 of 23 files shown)


tlambert03:
while you're in there... might also fix these `axes`

https://github.com/ome/ngff-spec/blob/36f74596b935606bfa53464ec18e808bce4b0531/examples/multiscales_strict/multiscales_example_relative.json#L9-L15

it's an object, rather than an array?

jo-mueller:
> it's an object, rather than an array?

Yeah that one was pretty weird. Fixed in e58c7c7702558ff05b32ebcb95b3ab3b497b9d38 👍

PR #170, opened 2026-07-02
fix: gh icon as relative path

jo-mueller:
Fix for the mildly annoying render of the authorlist over at this issue https://github.com/ome/ngff/pull/553

@lubianat @joshmoore

### DIFF
--- a/pre_build.py
+++ b/pre_build.py
@@ -178,7 +178,6 @@ def render_authors_md():
     # The invertocat icon was downloaded from https://brand.github.com/foundations/logo
     # and colored in grey according to the guidelines under the "Color" section.
     orcid_icon = "https://orcid.org/assets/vectors/orcid.logo.icon.svg"
-    github_icon = "icons/invertocat.svg"
 
     for idx, author in enumerate(authors):
         name = author['name']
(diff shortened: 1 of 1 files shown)


pre_build.py
```diff
@@ -187,7 +186,7 @@ def render_authors_md():
             name += f" [<img src=\"{orcid_icon}\" alt=\"ORCID iD\" height=12 width=12 style=\"vertical-align: middle;\"/>](https://orcid.org/{author['orcid']})"
 
         if "github" in author:
-            name += f" [<img src=\"{github_icon}\" alt=\"GitHub\" height=12 width=12 style=\"vertical-align: middle;\"/>](https://github.com/{author['github']})"
+            name += f" [<img src=icons/invertocat.svg alt=\"GitHub\" height=12 width=12 style=\"vertical-align: middle;\"/>](https://github.com/{author['github']})"
```
lubianat, line 189: will this actually fix it?  it seems that it does not find this path on the rendered version 

Image

maybe using https://github.githubassets.com/favicons/favicon.svg is safer? 

on ome/ngff#552 it looks like this to me: 

Image

lubianat (changes requested):

pre_build.py
jo-mueller: Over here this results in 
image

jo-mueller: which is why I went for the slightly altered, but stored in the repo version of the octocat.

> will this actually fix it?

I *hope* so, because the path to the icon was previously written as an absolute path (`/path...`), which worked in the ngff-spec repo, where root and relative path are the same. Hopefully turning it into a relative path will fix it though! But only one way to find out...

PR #173, opened 2026-07-09
Minor: Fix array coordinates example

btbest:
The example claims to show an array of shape (2, 3), but it actually shows an array of shape (3, 2) and also lists one out-of-range coordinate `(3, 1)`...

```
>>> numpy.ones((2, 3))
array([[1., 1., 1.],
       [1., 1., 1.]])
```

### DIFF
--- a/index.md
+++ b/index.md
@@ -298,9 +298,8 @@ The following conventions apply in this specification:
   the discrete coordinate system for that array defines the following array of points:
   ```
   [
-    [(0, 0), (0, 1)],
-    [(1, 0), (1, 1)],
-    [(2, 0), (3, 1)],
+    [(0, 0), (0, 1), (0, 2)],
+    [(1, 0), (1, 1), (1, 2)],
   ]
   ```
 - A "pixel"/"voxel" is the continuous region (rectangle/box) that corresponds to a single sample in the discrete array,


lubianat (commented):
thanks! (I think it is okay, but I don't know the context in depth, I'll wait for Johannes)

jo-mueller:
@btbest I think this one is actually incorrect. The example you are referencing does not refer to a list of points (which would be `(N, 2)` in 2D), but to an array of shape (2, 3). The indices necessary to index into that grid are the ones displayed in the backticks.

Maybe there *is* a better way of conveying this in markdown (maybe we need a figure?), but technically I think it's correct as it is.

btbest:
> The indices necessary to index into that grid are the ones displayed in the backticks.

A list of length 3 does *not* contain the index `3` with zero-based indexing as the example clearly uses. The example shows the index pair `(3, 1)` for an array of supposed shape (2, 3), when even just by that shape description, you can know that the largest index pair must be `(1, 2)` with zero-based indexing.

At the very least I hope you can believe me that the index in the first position has to be identical across any row, and the index in the second position has to be identical across any column. Which is not currently the case for the last row.

> The example you are referencing does not refer to a list of points (which would be `(N, 2)` in 2D)

As the text above the examples says: "For example, given an array with shape (2, 3), the discrete coordinate system for that array defines the following array of points:"

My PR doesn't change the example to an (N, 2) list of points. The example currently is a `List[List[Tuple[int, int]]]` and it remains a `List[List[Tuple[int, int]]]`.

I'm only fixing the number of entries and the last entry's coordinates.


I do think this example is so minor that it being wrong isn't particularly important, but ... it most definitely is wrong.

btbest:
Don't believe me, believe the LLM 🤣  https://chatgpt.com/s/t_6a71a3d3cb1081918735d939a317981d

jo-mueller:
Oh, I'm dumb. I was looking at the preview build and was like *Why, that's correct?*. 6 in the morning seems to be a bad time for PR review. Thanks!

jo-mueller (approved):

btbest:
Apparently also early enough for me to go full xkcd https://xkcd.com/386/ 😅  sorry about that

Slightly dumb of me to include an example in the PR description that isn't actually what I'm proposing to put in the text...

PR #175, opened 2026-07-29
Fix transformations examples

btbest:
Found after putting https://forum.image.sc/t/ngff-weekly-dev-update-thread/110810/129?u=btbest to the test and actually running all the examples through a roundtrip test :)

Edit: Just realised fixing that sequenceSubspace example makes it almost identical to the xarrayLike example, but oh well... it has a different child type 🤷

### DIFF
--- a/examples/transformations/byDimensionCoordinates.json
+++ b/examples/transformations/byDimensionCoordinates.json
@@ -11,21 +11,25 @@
   ],
   "coordinateTransformations" : [
     {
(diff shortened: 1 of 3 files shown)


jo-mueller (approved):

lubianat (approved):
I don't really know the details. But seems self-contained and I am trusting @jo-mueller. (adding the ok so he can merge)

PR #176, opened 2026-08-05
Remove strict jsonchemas and conformance test cases (v2)

### LINKED ISSUE #138: Remove "strict" jsonschemas and test cases

clbarnes:
Supercedes #140 ; resolves #138 .

Incorporates @will-moore 's feedback, migrating all of the strict examples to non-strict and updating references to them.


tests/test_validation.py
```diff
@@ -127,8 +122,8 @@ def test_example_configs():
...
     if missing:
-        raise Exception(f"Directories missing configs: {missing}")
+        raise FileNotFoundError(f"Directories missing configs: {missing}")
 
 
-if __name__ == '__main__':
-    pytest.main([__file__])
\ No newline at end of file
+if __name__ == "__main__":
+    pytest.main([__file__])
```
jo-mueller, line 129: Feel free to remove this alltogether, I think I left this behind at some point.

jo-mueller (commented):
I *think* it looks good. I'll briefly check the build on my end

tests/test_validation.py
clbarnes: The `if __name__ == "__main__"`? I think it can stay, not hurting anyone.

will-moore (approved):
Looks good to me, thanks 👍 

(not tested the dingus cli changes)

jo-mueller:
oh no, I somehow garbled up this PR while working on #188 🤦‍♂️

clbarnes:
@clbarnes to resolve merge conflicts (probably in another new branch)

clbarnes:
Included in #201

PR #183, opened 2026-08-11
General build configuration improvements

### LINKED ISSUE #182: Build system cleanup
There is some configuration which is a bit scattered; IMO it would be useful to centralise.

- [ ] Currently, the codespell configuration is split between pyproject.toml and .codespellrc. I recommend putting it in pyproject.toml.
- [ ] Currently, codespell checks a number of directories it shouldn't, including .venv, .tox, and _build; we should add these to the ignore section of the config.
- [ ] Currently, tox is used to manage an environment in which to run pytest. However, uv can do this for us; tox would be useful if we were running jobs across multiple environments, which we're not. I recommend scrapping tox entirely and just using `uv run pytest`.
- [ ] Currently, commands for common dev tasks (building, linting, testing etc.) are tool-specific and scattered in various config files. I propose putting them all in a [justfile](https://github.com/casey/just). The only downside of just is that it isn't easily available on readthedocs. The downside of make is that using make as a task runner is sort of an abuse of what make is for.

clbarnes:
Resolves #182 

- Consolidate codespell config into pyproject.toml (previously split between pyproject.toml and .codespellrc)
- Remove tox, just use `uv run pytest` as we only had 1 tox env anyway
- Add common dev tasks to a justfile (`just test`, `just build`, `just lint` etc)


pyproject.toml
```diff
@@ -15,12 +15,14 @@ dependencies = [
 
 [project.optional-dependencies]
 testing = [
-    "tox",
+    "codespell",
     "pytest",  # https://docs.pytest.org/en/latest/contents.html
```
jo-mueller, line 19: If `just` is the way to run all CI checks here, then it should be part of the deps

README.md
```diff
@@ -9,6 +9,11 @@ as well as [json schemas](./schemas) to validate written ome-zarr image data.
 
 The built documentation including contribution hints can be found **[here](https://ngff-spec.readthedocs.io/en/latest/)**.
 
+## Development
+
+Use [`just`](https://github.com/casey/just) for common development tasks like building, linting, testing.
+Run `just` to see a list of commands.
```
jo-mueller, line 15: Let's add the actual command to run/build stuff here. These commands are also mentioned in the contributing section of the repo, so that'll have to be updated there, too.

jo-mueller (commented):
Looks alright to me. I'm a bit agnostic about the justfile, but just because it's not super common in the Python ecosystem, but that's probably just my bubble. If we add it, installation and usage should be spelled out a bit more clearly.

clbarnes:
@clbarnes to add just to contributing file and pyproject.toml

pyproject.toml
clbarnes: Done (called rust-just)

README.md
clbarnes: Updated in contribute.md. I would prefer not to document all the recipes here to reduce duplication/ potential for drift. The existing docs tells the user how to access the list themselves.

clbarnes:
Included in #201

PR #190, opened 2026-08-19
Update spec for 0.9.dev1 including RFCs 3 & 4

jni:
0.9.dev1 is intended to be:
- 0.6rc1
- with RFC3
- and RFC4
- and ~~probably~~ #87 

in current WIP this only includes RFC3. I need to get RFC4 in before we merge this.

**Edit 1 (2026-08-20):** updated title to 0.9.dev1 to reflect discussion and match PEP-440.

**Edit 2 (2026-08-21):** This PR now contains all of the listed requirements (RFC3, RFC4, #87 omero fields optional) and should be ready to merge to allow implementations to start to be written against it. Some notes below.

## notes

The limits removals on coordinate-transformations are the thing I'm least sure about, would appreciate extra eyes there.

In `schemas/coordinate_transformations.schema`, `rotation` is defined by a 2, 3, 4, or 5-dimensional rotation matrix. This is incompatible with RFC-3 but cannot be easily extended to an arbitrary number of dimensions.

Possible solutions:
- Amend RFC-3 to have a large number of max dimensions (say, 12? 16?), and continue the definition given the arbitrary rotations
- For higher-dimensional rotations, allow instead a sequence of 2-axis rotations, which compose into a high-dimensional rotation. (Can higher-dimensional rotations always be decomposed in this way?)
- Amend 0.6/RFC-5 to disallow >3D rotations, since they are very uncommon, and can already be decomposed (as a sequence of 2D rotations, because sequences are already a thing.)

I'm also not sure about 6199f0e and the difference to 93c1ba7.

Finally I don't know how to express that orientation can only be present if axis type is "space".


sbesson:
@jni thanks for starting this work. We're looking forward to having a first rendered version of this specification as well as dev schemas that we can use to validate datasets generated by implementations!

Immediate question: is the next version number 0.9dev1 (as in this PR) or 0.9.dev1 (used in a few public discussions)? 

We have no opinion one way or the other, but we have used the latter so far so if the consensus is to move towards the first form, we will need to update our implementation and regenerate sample data /cc @melissalinkert

clbarnes:
> is the next version number 0.9dev1 (as in this PR) or 0.9.dev1 (used in a few public discussions)

This came up with the dev versions in 0.6. Because we use python tooling in the build step of the repo, we normalised on [PEP440-style version specifiers](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers) `0.6.dev0` and `0.6rc0`. Overfitting to the language we happen to be using for the build is not ideal, but it's pretty well specified with implementations in many languages, while allowing a lot of flexibility in the information included in a version string. The inconsistency in `.` between the dev and a/b/rc is annoying, but I suppose it's intended so that you can have a dev version for an alpha release, and so on. [This rust crate](https://crates.io/crates/pep440_rs) has a good run-down of the downsides of PEP440.

Standard [SemVer](https://semver.org/) would use something like `0.9-dev.1` but, other than `.`-delimiting, treats string and numeric components of the pre-release version as arbitrary, to be sorted lexicographically and numerically respectively. e.g. `dev` would come after `alpha` and `beta`. It also doesn't allow post-releases (probably fine, a post-release should probably be a patch release) or epochs e.g. if we decided we wanted to switch to calver later (also probably fine).

IMO we should stick with PEP440 (`0.9.dev1`), or switch to standard SemVer (`0.9-dev.1`) and separate the versioning of the project from the versioning of the python tooling used to build it, but I don't care too much about which.

melissalinkert:
Just to be clear, then, is the use of `0.9dev1` now fully decided? I need to update https://github.com/glencoesoftware/bioformats2raw/pull/330 and https://github.com/glencoesoftware/raw2ometiff/pull/156 with the agreed-upon version number before merging and tagging release candidates, which we expect to do within the next few days. If the version number is still being discussed, I need to know so that merging/tagging can be delayed until the discussion is resolved.

jni:
Apologies, for the confusion, my confusion indeed stemmed from the inconsistency between 0.6.devN and 0.6rcN noted by @clbarnes above. I think consistency with the 0.6 series makes sense, which means I'll update this PR to match that. In short @melissalinkert @sbesson, let's go with 0.9.dev1, no need to change your outputs, I will update this PR accordingly.

jni:
Here's the current validation test failures:

```
FAILED tests/test_attributes.py::test_attributes[spec/invalid/image/missing_space_axes] - Failed: DID NOT RAISE ValidationError
FAILED tests/test_attributes.py::test_attributes[spec/invalid/image/non_existent_orientation] - Failed: DID NOT RAISE ValidationError
FAILED tests/test_attributes.py::test_attributes[spec/invalid/image/non_existent_orientation_type] - Failed: DID NOT RAISE ValidationError
FAILED tests/test_attributes.py::test_attributes[spec/invalid/transforms/bad_projectAxis_insert_too_high_dim] - Failed: DID NOT RAISE ValidationError
FAILED tests/test_attributes.py::test_attributes[spec/invalid/transforms/bad_projectAxis_insert_too_many] - Failed: DID NOT RAISE ValidationError
FAILED tests/test_attributes.py::test_attributes[spec/invalid/transforms/bad_projectAxis_remove_too_high_dim] - Failed: DID NOT RAISE ValidationError
FAILED tests/test_attributes.py::test_attributes[spec/invalid/transforms/bad_projectAxis_remove_too_many] - Failed: DID NOT RAISE ValidationError
FAILED tests/test_attributes.py::test_attributes[spec/valid/image/many_space_axes] - jsonschema.exceptions.ValidationError: {'ome': {'version': '0.9.dev1...
```

Most of these are because they have now become valid, so I'll move them. But these two:

```
FAILED tests/test_attributes.py::test_attributes[spec/invalid/image/non_existent_orientation] - Failed: DID NOT RAISE ValidationError
FAILED tests/test_attributes.py::test_attributes[spec/invalid/image/non_existent_orientation_type] - Failed: DID NOT RAISE ValidationError
```

I don't understand why they are failing. Here's the orientation schema, which specifies that the strings must come from specific lists:

https://github.com/jni/ngff-spec/blob/af560df45c0aa987d537e128195bc2dcd454c0a4/schemas/axis_orientation.schema#L11-L43

and here's the relevant part of the files that are unexpectedly passing validation:

https://github.com/jni/ngff-spec/blob/af560df45c0aa987d537e128195bc2dcd454c0a4/tests/attributes/spec/invalid/image/non_existent_orientation.json#L16-L18

https://github.com/jni/ngff-spec/blob/af560df45c0aa987d537e128195bc2dcd454c0a4/tests/attributes/spec/invalid/image/non_existent_orientation_type.json#L26-L29

I'm a noob with jsonschema so I would appreciate any tips or advice here!

jni:
@joshmoore @clbarnes @jo-mueller This should be ready for review. I've also rebased my changes into logical commits to make review easier. Especially if you skip the last commit (which updates the version string in every file), the list of changes should be manageable.

tests/test_attributes.py
```diff
@@ -30,7 +30,18 @@
 
 case_fnames = sorted(attrs_dir.rglob("*.json"))
 
-xfails = set()
+xfails = set([
```
clbarnes: Extremely minor point: can use a set literal `{}` here.

index.md
```diff
@@ -266,10 +265,56 @@ where each object describes a dimension (axis) and:
 - MAY contain the field `longName`.
   The value MUST be a string,
   and can provide a longer name or description of an axis and its properties.
+- MAY contain an "orientation" field, itself containing a "type" and a
+  "value". Currently, "type" may only be "anatomical", and "value" may
+  only be one from the list given in [list of possible anatomical
+  orientation values](orientation-values). A specific orientation value or
+  its inverse MUST NOT appear more than once in a given coordinate system.
```
clbarnes, line 271: Could RECOMMEND TCZYX here for backwards-compatibility, if there's any benefit to that? Or alternatively, add a `{hint}` that axis ordering could affect performance of certain quries in combination with configuration of the array chunking/ sharding/ transposing.

index.md
```diff
@@ -1295,15 +1340,26 @@ Each object provides the following fields:
...
-  - The length of `axes` must be between 2 and 5 and MUST be equal to the dimensionality of the Zarr arrays storing the image data (see `datasets:path`).
-  - `axes` MUST contain 2 or 3 entries of `type:space`
-  - `axes` MAY contain one additional entry of `type:time`
-  - `axes` MAY contain one additional entry of `type:channel` or a null / custom type.
-  - `axes` entries MUST be ordered by `type` where the `time` axis must come first (if present),
-    followed by the  `channel` or custom axis (if present) and the axes of type `space`.
-  - If there are three spatial axes where two correspond to the image plane (`yx`)
-    and images are stacked along the other (anisotropic) axis (`z`),
-    the spatial axes SHOULD be ordered as `zyx`.
+0. The length of the axis names MUST match the number of axes of the array.
```
clbarnes: I'd be inclined towards 1-basing for ordered lists. The audience is obviously technical but in text 1-based is much more common.

clbarnes (approved):

index.md
sbesson: This line should probably be updated following RFC-3

index.md
jni: @clbarnes I'm puzzled about the line you chose to make this comment on? I don't see why axis an orientation note would recommend anything about axis ordering?

At any rate, I think hints about performance can come in future iterations of the 0.9.devN series.

index.md
jni: Oh, I don't mind fixing that, I didn't because I added it later and it renders as 1-based anyway:

image

So I suggest fixing that in later refinement passes also.

jni: ... But the wrong tick marks in point 5 should definitely be fixed :joy:

jni: ok I fixed both the numbering and the quotes.

index.md
jni: fixed! :pray:

index.md
clbarnes: I am also puzzled by my choice of line here. Maybe I fat-fingered the touchpad.

joshmoore:
Kudos, @jni! ❤️ For everyone working towards OME-Zarr 1.0, this will be the first of many dev tags. `dev2` is likely to be a clean up tag short on the heels of dev1, led by @clbarnes. If there's something you are looking to get in, please get in touch (open an issue, ask on Zulip, etc. etc) to get your very own devX. :smile:

sbesson:
Adding a late 👍 following a discussion yesterday at the weekly OME Formats meeting.

From a consumer perspective, are we expecting the development specification to be published under https://ngff.openmicroscopy.org/specifications/dev/index.html or https://ngff.openmicroscopy.org/specifications/0.9.dev1/index.html? Is that possibly waiting on `0.6.0` to be fully released?

jni:
dev/ is waiting for 0.6.0 to be released (there can be only one 👨🏴󠁧󠁢󠁳󠁣󠁴󠁿🌩️). 0.9.dev1/ *should* be served shortly, see ome/ngff#581 for some discussion.

will-moore:
Just looking at https://github.com/ome/ngff-spec/releases/tag/0.9.dev1 it's a shame that page doesn't mention that RFC3 and RFC4 are part of 0.9dev1.
That could possibly be improved a bit by renaming this PR to include RFC3 and RFC4 in the name? Hopefully that will then appear on the 0.9.dev1 page, even if just within the listed PRs.

jni:
I'm not sure but I don't believe updating the text here will update it on the release page. Still, I'll do it for future releases.

We can also write a more thorough description for 0.9.dev2 which we can tag reasonably quickly.

PR #197, opened 2026-08-28
chore: fix broken cross-reference

jo-mueller:
Just an internal cross-reference broken, which was flagged by the linter over at https://github.com/ome/ngff/pull/581


clbarnes:
Included in #201

PR #198, opened 2026-08-28
chore: fix anchors and references

jo-mueller:
Copy of #197 plus another broken cross-ref that appeared over at https://github.com/ome/ngff/pull/581


clbarnes (approved):

clbarnes:
Included in #201

PR #199, opened 2026-08-28
Change input name from 'intrinsic' to 'physical'

clbarnes:
Intrinsic is not a defined coordinate system.


jo-mueller (approved):

lubianat (approved):
I don't have the detailed bg, but LGTM

Issue #191, opened 2026-08-19
Format object definitions as tables


clbarnes:
Reading and writing prose is not very accessible for well-defined structural information like defining data structures.
Putting them in tables would be much easier for implementors, much more referenceable, easier to get an overview, etc..
In my opinion, the definitions in tables should be normative.

## Example

I applied this to part of the NGFF v0.5 spec [in a separate repo](https://github.com/clbarnes/ngff2/blob/main/SPECIFICATION.md#object-specifications). Excerpt:

```markdown
### Object: Multiscale

| key | necessity | type | description |
| --- | --------- | ---- | ----------- |
| `axes` | MUST | array of [Axis](#object-axis) | Axis information for this multiscale's coordinate system. |
| `datasets` | MUST | array of [Dataset](#object-dataset) | Single-scale datasets composing this multiscale dataset |
| `name` | SHOULD | string | Name of the multiscale dataset |
| `type` | SHOULD | string | Type of downscaling method |
| `metadata` | SHOULD | object | Additional information about the downscaling method and a new line |
| `coordinateTransformations` | MAY | array of [Coordinate Transformation](#tagged-union-coordinate-transformation) | Transformations applied in sequence from some common coordinate system into physical coordinates |

#### Multiscale field: `datasets`

- MUST be ordered highest-resolution (largest) to lowest-resolution (smallest)

#### Multiscale field: `axes`

- MUST have the same length as the dimensionality of all Zarr arrays in this multiscale
- Axis objects must have unique `name` fields in this metadata array
- All Zarr arrays MUST have the `dimension_names` field set, which MUST match the order and values of the `name` fields of this `axes` array
- MUST contain between 2 and 5 elements, with types which MUST be in the order
  - optional axis with [`type`](#enum-axis-type) `"time"`
  - optional axis with [`type`](#enum-axis-type) `"channel"` (conventionally with `name` `"c"`), OR a custom string `type`, OR `null`, or omitted
  - optional axis with [`type`](#enum-axis-type) `"space"`, conventionally the anisotropic slicing axis if appropriate, conventionally with `name` `"z"`
  - required axis with [`type`](#enum-axis-type) `"space"`, conventionally with `name` `"y"`
  - required axis with [`type`](#enum-axis-type) `"space"`, conventionally with `name` `"x"`

#### Multiscale field: `coordinateTransformations`

[...]

Issue #184, opened 2026-08-12
Reorder spec sections so the direct readability is improved 


lubianat:
### Describe the issue

https://ngff.openmicroscopy.org/specifications/dev/index.html#coordinatetransformations-metadata is a huge block, I wonder if it could be moved below?

Parts look semi-alphabetically ordered, not sure on how ordering was picked. But current order made a lot of sense for 0.5, where "“coordinateTransformations” metadata" was small

<img width="189" height="925" alt="Image" src="https://github.com/user-attachments/assets/c29847ce-125a-42b3-a5a7-0209756ab4a0" />

now it feels a bit weird to have `multiscales` (arguably a core part, common to all OME-Zarrs) come after all the different transform types

update: 

– Added a PR to move multiscales to the top, it makes it more readable IMO

PR #201, opened 2026-09-02
0.9.dev2: non-feature improvements

### LINKED ISSUE #123: Pre-commit hooks for enforcing formatting
[pre-commit](https://pre-commit.com/) and its RiiR sibling [prek](https://prek.j178.dev/) manage a bunch of different pre-commit hooks which could make development more consistent. They are broadly compatible, with prek being faster and more efficient for most common actions.
Using prek, I tried

- trailing-whitespace
- end-of-file-fixer
- check-yaml
- check-json
- pretty-format-json
- mixed-line-ending, fix=lf
- proselint
- typos
- codespell

Most are from [prek's builtins](https://prek.j178.dev/builtin/#2-explicit-builtin-repository); a few are from [pre-commit's recommendations](https://pre-commit.com/hooks.html).

Any of them will introduce quite a lot of non-meaningful git churn and so should be added immediately after a release, ideally one at a time.

In particular, end-of-file-fixer and pretty-format-json edit a lot of the JSON examples.

proselint helped me fix some issues but gets a few false positives on helpful lints, so we may not be able to use it here.
codespell didn't find anything useful.
typos found some false positives which could be configured around (e.g. flagging "OME" as a misspelling of "some"), but then didn't find some other actual typos.
These should probably be omitted.

### LINKED ISSUE #138: Remove "strict" jsonschemas and test cases

### LINKED ISSUE #194: lint/validate JSONs in the repository
### Describe the issue

From 
- https://github.com/ome/ngff-spec/pull/193

it would be good to catch those automatically

clbarnes:
This dev release is intended to improve the contributor and possibly reader experience without introducing any normative changes to the spec itself.

The nature of this dev version means there look like a _lot_ of changes to review. Most of this is fixing many years' worth of accrued whitespace errors and inconsistent formatting. This PR adds config to make sure we never have to worry about this again, so it's worth the one-time pain to get that config in place. 

## Goals

- [x] Keep in sync with recent changes to `main`
- [x] resolve #138 
- Add additional lints and some auto-format jobs to improve consistency end reduce e.g. whitespace-only git churn
  - [x] resolve #123
  - [x] resolve #194

### Stretch

- [ ] #191
  - This is currently implemented as a PR on top of this one, at https://github.com/clbarnes/ngff-spec/pull/3 

## Catalogue of changes

Because the diff is nightmarish.

Incorporating recent fixes to main/ 0.9.dev:

- Schemas allow the minor version as well as prereleases of it (first introduced in v0.6) https://github.com/ome/ngff-spec/pull/201/commits/d4b2c1dc35e8affa4667d558599c46ecd812dcff
- Fix some JSON typos from 0.9.dev1: https://github.com/ome/ngff-spec/pull/201/commits/5c76733fe3a1f97a3957909c879b15d57689f74e
- describe byDimension input/outputAxes as indices, not names, and fix image schema typo: https://github.com/ome/ngff-spec/pull/201/commits/eb227a48cd8260d23ffe7c8db7c85c8d3a4f1bd9
- bump jupyter-book dependency to 2.0.2 per https://github.com/ome/ngff-spec/pull/204
- manually apply changes from 
  - https://github.com/ome/ngff-spec/pull/202
  - https://github.com/ome/ngff-spec/pull/203
  - https://github.com/ome/ngff-spec/pull/205
  - https://github.com/ome/ngff-spec/pull/211

Lint and CI configuration:

- Previously, codespell configuration was split between .codespellrc and pyproject.toml; now it's all in pyproject.toml. Also, added codespell to the dependencies https://github.com/ome/ngff-spec/pull/201/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R33
- Minor .editorconfig changes to better suit our use of certain file formats https://github.com/ome/ngff-spec/pull/201/changes#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5bR9
- [prek](https://prek.j178.dev/) configuration for running pre-commit hooks (similar to [pre-commit](https://pre-commit.com/) and compatible with most of the same recipes, but much faster) https://github.com/ome/ngff-spec/pull/201/changes#diff-72359463e1c50ddcb8dfabe900898a8f1d92416bbe6754f2e497f62d96232021 The configured hooks are
  - remove trailing whitespace
  - ensure every text file has a newline at the end of it
  - ensure that LF is used as the line ending
  - check that YAML files are valid
  - check that JSON files are valid
  - format JSON files consistently
  - check that all files are consistent with the .editorconfig
  - run codespell
- document use of pre-commit hooks https://github.com/ome/ngff-spec/pull/201/changes#diff-0e6105c1be2a3256539260ca3c7fe14f58a845005496d30e7e897cf9cf9c65e8R50
- run pre-commit hooks on CI https://github.com/ome/ngff-spec/pull/201/changes#diff-c707f934fe56964ebe4e68cbf0eb00915fb01d8af63d6b37a06dbf365e9cf937R16
- use uv and just throughout the CI configuration (see `.github/workflows/*.yml`)
- add a CI job which actually builds the site and could run lints on the built artefact https://github.com/ome/ngff-spec/pull/201/changes#diff-3c3f6dcb27eee72f5e9c8e081a7487447b247fe8a7097d1bd973bc180fd0869dR1
  - I wanted to use htmlproofer here, as ome/ngff does, but it ran for half an hour and produced thousands of lines of output which wasn't very helpful
- Run CI jobs on every push/PR

Schema simplification

- Remove strict schemas and test cases as discussed in https://github.com/ome/ngff-spec/issues/138
- Rename the "strict" examples (included in the document - see examples/*_strict

Environment management

- Previously, a "testing" extra was used to install test-only dependencies. While this isn't a python package so the usual distinction between runtime and development dependencies is not so clear, it is preferred today to use a [dependency group](https://packaging.python.org/en/latest/specifications/dependency-groups/), which is implemented here https://github.com/ome/ngff-spec/pull/201/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R22
- Previously, pip was recommended for package management, environment management was left unspecified, there was no lock file, and tox was used for controlling the test environment. This, plus installing python itself, is now managed by uv, which should also speed up build times on CI https://github.com/ome/ngff-spec/pull/201/changes#diff-0e6105c1be2a3256539260ca3c7fe14f58a845005496d30e7e897cf9cf9c65e8R24

Build scripts

- Previously, common tasks like building or viewing the docs, generating the citations file, and running the tests required different commands which were scattered across different documentation and configuration files. Contributors can now use [just](https://github.com/casey/just) (whose normal usage is very similar to how Make is abused in python development); the justfile makes for a nice self-documenting catalogue of those commands https://github.com/ome/ngff-spec/pull/201/changes#diff-deb9bb56fb122db0b605aa5b63f95a4665c905b18dd670e1fa6c877576a94ff1
  - installation of just is managed by uv

Minor changes in major volume

- update 0.9.dev1 to 0.9.dev2
- fix whitespace and minor formatting issues to suit the new editorconfig/ pre-commit configuration
- some very minor formatting in python scripts (mainly sorting imports)


jni:
Thanks @clbarnes! Can you add more details about the changes due to the large number of files?

>  Keep in sync with recent changes to main

which recent changes? Can you list them? Is this all a straight back/forward port or have you made the changes here optimistically, expecting that dev will become main soon?

> Add additional lints and some auto-format jobs to improve consistency end reduce e.g. whitespace-only git churn 

Can you list the configuration, or link to the specific files to review?

Basically, it would be good if this PR had a good record of exactly the changes that were made, since the diff itself is enormous and thus difficult to review.

Otherwise, full support! :rocket:

clbarnes:
Good feedback @jni ; I've added a change catalogue to the top post.

clbarnes:
I got nerdsniped into [adding JSONC support to prek](https://github.com/j178/prek/pull/2682); not sure if it'll be accepted, but it would allow us to check that the examples are valid JSON(C).

clbarnes:
Object definitions as tables where fields' types are linked to their own definition gives you a nice hover preview. From https://ngff-spec-clbarnes.readthedocs.io/en/0.9.dev2-tables

Screenshot 2026-09-03 at 17 13 33

lubianat:
@clbarnes would you mind adding also 
- https://github.com/ome/ngff-spec/issues/184 
as many things are being touched? it can also be done later, though

it could be a PR on top of a PR on top of a PR...

clbarnes:
@lubianat and I discussed in person but for posterity, IMO doing significant rearranging of the spec is too much intervention for this dev release, which I wanted to be just about firming up the build/ lint setup and then strictly adding the tables, both of which I expect to be much less controversial. I suspect that as 0.9 grows we'll need to do at least another round of rearranging so it could go in there.

On https://github.com/ome/ngff-spec/pull/201#issuecomment-5525396338 : prek has merged my JSONC hook so it should be in the next release, but we shouldn't block this PR on that.

lubianat:
@clbarnes any suggestions on how to review this? also, should we let Copilot Lite run and look for typos or similar?

clbarnes:
I would recommend going through the catalogue of changes in the first issue to see if you agree with the concept. That catalogue mainly points out which files have changed, but if you approximated with "everything which isn't a test, example, schema, or markdown" that would get most of it. I'd be concerned about copilot throwing out a lot of false positives on issues unrelated to my changes; then again, maybe now is the time to fix any such issues.

This PR shouldn't have any meaningful changes to the spec itself other than whitespace and a couple of renames where the strict examples have been renamed; the table PR is much more reviewable.

jni (approved):

jni:
I'm personally very excited about this change after having dealt with 0.9.dev1 😂 Thank you @clbarnes!

clbarnes:
Test failures are due to https://github.com/ome/ngff-spec/issues/212 , i.e. fixing a bug to actually run some tests we thought we were running but which are very out of date and possibly not worth keeping.

schemas/_version.schema
```diff
@@ -1,10 +1,11 @@
...
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://ngff.openmicroscopy.org/0.9.dev1/schemas/_version.schema",
+  "$id": "https://ngff.openmicroscopy.org/0.9.dev2/schemas/_version.schema",
   "title": "OME-Zarr version",
   "description": "OME-Zarr version.",
   "type": "string",
   "enum": [
-    "0.9.dev1"
+    "0.9.dev2",
+    "0.9"
```
jo-mueller, line 9: Do you want to allow the version 0.9 to be a valid version key? I think for 0.9 this should be handled a bit tighter. In 0.6 and 0.6rc0 it was ok because we communicated that no fields would change from 0.6rc0 -> 0.6 so 0.6 would be backwards-compatible with 0.6rc0.

It's unlikely that this is the case with 0.9dev versions. If anything you could put `0.9dev1` and `0.9dev2` here, which would signal that these two are specification-compatible.

schemas/scene.schema
jo-mueller, line 17: Not sure whether it belongs in this PR but we may want to use the full ref (`https://...coordinate_systems.schema`) here as well. But that can be its own PR, too.

jo-mueller (approved):

lubianat:
how do you want to handle https://github.com/clbarnes/ngff-spec/pull/3? should that be merged first or we just go ahead here? 

with Johannes and Juan ok'ing, I'll also ok and leave @clbarnes to merge when ready

.github/workflows/build-lint.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - uses: actions/checkout@v7
+      - uses: astral-sh/setup-uv@v9.0.0
+      - uses: extractions/setup-just@v4
+      - run: just build
+      # - name: Validate all links, enforce alt text
+      #   uses: athackst/htmlproofer-action@main
+      #   with:
+      #     directory: _build/html
+      #     disable_external: true
```
lubianat, line 24: This is commented out; should we remove? Or a separate `workflow_dispatch` action?

uv.lock
```diff
@@ -0,0 +1,2105 @@
+version = 1
+revision = 3
```
lubianat, line 2: I think you did commit the lock on purpose, right? Fair enough.

lubianat (approved):

uv.lock
clbarnes: Yes; as ngff-spec is a document which we want to build reproducibly rather than a library we want to work in a range of host environments, I think the lock file is a positive here.

.github/workflows/build-lint.yml
clbarnes: Conceptually I'd like to run htmlproofer but running this took ages and produced thousands of errors. Probably best to remove and stick an issue on the backlog.

jo-mueller: The reason for that is likely the markdown pages produced by the schema renderer. But it's probably easier to switch to linkML instead of fixing this.

jni:
@clbarnes @jo-mueller I remember there was an issue with incorrect schema references (to `axes.schema`, iirc) in 0.9.dev1, but I can't find the issue. Could either of you point me to it *and* tell me whether this PR fixes it? Or am I just going crazy :pray:

clbarnes:
Was it this one? Linked in the PR description, merged into main and also implemented here https://github.com/ome/ngff-spec/pull/203

jo-mueller:
Note that #203 only changed that for axes.schema, but we should do the same for coordinate_systems.schema and coordinate_transformations.schema consistently.

Doesn't have to be this PR though.

clbarnes:
Note that I have merged the object tables, as well as making the remainder of the cross-schema references absolute IRIs.

PR #202, opened 2026-09-04
bump version to 0.6

jo-mueller:
Title - bumps all the versions to 0.6 before the next release :)


version_history.md
```diff
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 0.6rc0 - 2026-07-01
```
clbarnes, line 11: The date on this probably needs to change too

jo-mueller: 59b01888f329e175c63c0eff6ac1cd8a886561e1

jo-mueller:
ugh, versions got garbled again the bump. Will need to have a more serious look at #188 soon!

uv.lock
```diff
@@ -0,0 +1,2265 @@
+version = 1
+revision = 3
+requires-python = ">=3.10"
+resolution-markers = [
+    "python_full_version >= '3.14'",
+    "python_full_version >= '3.11' and python_full_version < '3.14'",
```
lubianat: was the uv.lock commit intentional?

jo-mueller: no :)

clbarnes: As this is not a library which needs to play nice with python environments containing other packages, but rather a document which needs to be built reproducibly, I'd be inclined towards keeping the lock file. Unless it interferes with the ngff build process?

lubianat: I don't particularly care, but I do think it should not be just hidden in this PR he 

at least as a separate commit (or a separate mini PR)

jo-mueller: 9503de8e311dce851d0b28964929c1ecc27a63ea

lubianat (approved):

joshmoore (approved):
:tada:

PR #203, opened 2026-09-07
schema: fix reference to axes to use URLs, too

jo-mueller:
Pointed out by @will-moore more [here](https://github.com/ome/ngff/pull/581#issuecomment-5496615135). The reference to the axes schema doesn't use the correct format because `axes` are not defined in the same schema document. This likely will need to be fixed for 0.1 .... 0.5 as well.

*Edit*: The axes schema doesn't sit in its own document for version 0.1...0.5, so this doesn't need to be backported 👍


schemas/coordinate_systems.schema
```diff
@@ -19,7 +19,7 @@
           "description": "Name of coordinate system. Must be unique among all coordinate systems."
         },
         "axes": {
-          "$ref": "axes.schema"
+          "$ref": "https://ngff.openmicroscopy.org/0.6/schemas/axes.schema"
```
will-moore: Other schemas currently use `0.6rc0` e.g. 

```
"$ref": "https://ngff.openmicroscopy.org/0.6rc0/schemas/coordinate_transformations.schema#/$defs/inputOutput"
```

but I'm not sure that explains the 403 test errors for https://ngff.openmicroscopy.org/0.6/schemas/axes.schema
since that and https://ngff.openmicroscopy.org/0.6rc0/schemas/axes.schema are 404 in the browser?

jo-mueller: Right. Change was made over at #202 originally, so it was pointing at a different version. This should then go in before that PR. (29b1f5afa20ac4ce86fe7ac29dd3989dc3f2522c)

jo-mueller:
I was also just looking at the 0.5 branch, where `axes` don't exist in their own separate file, so it doesn't need to be referenced via an URL. I think this would be the only place where this fix needs to be made. 

In fact, I don't see a compelling reason why the axes.schema has to be its own file (that a schema resolver would need to find) in the first place. `axes` metadata is only ever used inside `coordinateSystems` metadata. This design choice may have historic reasons (@bogovicj ?). 

If nothing speaks against it, I could inline the axes.schema into the coordinate_systems.schema in a follow-up PR. Unless I end up breaking the validator or something like that

gouttegd (approved):

will-moore:
@jo-mueller Currently, the validator will try to load a `axes.schema` for v0.6 since it's hard-coded.
But with https://github.com/ome/ome-ngff-validator/pull/86/changes we detect find which schemas need loading dynamically. So it would be nice to get that in (which depends on https://github.com/ome/ngff/pull/581) before inlining the axes.schema.

lubianat (approved):
I don't have the complete background, but it is self contained and I trust @jo-mueller with  the decision.

will-moore:
Thanks, this is fixed for v0.6.

It's still unchanged for v0.6.dev4 at https://raw.githubusercontent.com/ome/ngff-spec/0.6.dev4/schemas/coordinate_systems.schema but that's kinda minor issue now.
I'll leave the workaround in for that at https://github.com/ome/ome-ngff-validator/pull/86 for now, which is OK.

will-moore:
https://deploy-preview-86--ome-ngff-validator.netlify.app/?source=https://radosgw.public.os.wwu.de/rfc5-transform-test-data/P2A_B6_M2_Scene_2.ome.zarr/ is giving an error: 
```
can't resolve reference coordinate_systems.schema from id https://ngff.openmicroscopy.org/0.6/schemas/scene.schema
```

Which I haven't fully debugged, but I think is due to the different ways that schemas are referenced in
https://raw.githubusercontent.com/ome/ngff-spec/0.6/schemas/scene.schema 

This has various refs:

```
"$ref": "https://ngff.openmicroscopy.org/0.6/schemas/_version.schema"
"$ref": "coordinate_systems.schema"
"$ref": "coordinate_transformations.schema#/$defs/coordinateTransformation",
```

The last 2 should probably look like:

```
"$ref": "https://ngff.openmicroscopy.org/0.6/schemas/_version.schema"
"$ref": "https://ngff.openmicroscopy.org/0.6/schemas/coordinate_systems.schema#/$defs/coordinateSystem"
"$ref": "https://ngff.openmicroscopy.org/0.6/schemas/coordinate_transformations.schema#/$defs/coordinateTransformation",
```

jo-mueller:
I think this should be fixed once https://github.com/ome/ngff/pull/612 is merged. Currently, I think there's nothing on the ngff page that serves 0.6 schemas, because version 0.6 is available neither as a tag or a submodule.

- Tag: 0.6 *was* tagged, but the page hasn't been rebuilt since then so the respective schemas aren't served accordingly
- Submodule: See https://github.com/ome/ngff/pull/612. This adds 0.6 as a submodule, the schemas should then be hardcopied to the ngff page's `_html_extra` from where the schemas are served.

will-moore:
OK, happy to wait and see, but even the raw schema https://github.com/ome/ngff-spec/blob/main/schemas/scene.schema has these different refs, similar to what was fixed in this PR above.
So it looks like they need the same fix?

PR #205, opened 2026-09-11
Prepend anchors 0.6

jo-mueller:
The anchors that are autogenerated in the `pre_build.py` script (i.e., [here](https://github.com/ome/ngff-spec/blob/cf12979769b0322ed21575889823025be350bd7f/pre_build.py#L116) on main) are generated in exactly the same way on all the other branches of ngff-spec, which leads to confusion with how anchors are resolved over at ngff (see also https://github.com/ome/ngff/pull/607)

This solves it on main, but the fix will have to be ported over to the 0.1 ... 0.5 branches, too.

Another small issue I noted - the examples section lists its own index.md as part of the examples:

image

Which is fixed in ee4b16806d96d4b79d4a3b9eed79c6f6abb55eb5 here, too. Big screenshot for a small fix.

### DIFF
--- a/myst.yml
+++ b/myst.yml
@@ -97,7 +97,7 @@ project:
     - file: index.md
     - file: examples/index.md
       children:
-        - pattern: examples/**/*.md
+        - pattern: examples/*/*.md
     - file: schemas/index.md
       children:
         - pattern: schemas/*.md
--- a/pre_build.py
+++ b/pre_build.py
@@ -4,6 +4,7 @@ import glob
 from pathlib import Path
 import jsonc as json
 import logging
+from _version import __version__
 
 # Suppress warnings from json-schema-for-humans about unresolvable URLs
 logging.getLogger().setLevel(logging.ERROR)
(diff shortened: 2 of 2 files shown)


jo-mueller:
@lubianat @clbarnes I think this would actually be important to get in before tagging 0.6 as part of the fix for https://github.com/ome/ngff/pull/607. After 0.6 is tagged, we cannot change it because we cannot edit the tag.

We could (and probably should) archive 0.6 on a different branch though, so we can make these small build fixes for 0.6 in the future, too.

clbarnes (approved):

clbarnes:
One thing to consider about the anchors is that the colons get normalised to hyphens in the built site, so maybe it's best to write them as hyphens in the first place to minimise confusion/ in case someone tries to differentiate two anchors by colon/hyphen?

lubianat (approved):

lubianat:
I approved it in all to allow merge and leave @jo-mueller to decide what to do with @clbarnes's comment (which makes sense) 

I think though we should stop supporting changes on 0.1, 0.2 and 0.3 (unless they are security vulnerabilities)....

lubianat (approved):

